PR #40, opened 2022-01-17
Prevent adding non-"character" actors to party

freohr:
The party display filters out non-"characters", but using the "party" could be useful outside of the party sheet, so non-characters (only monsters for now) actors cannot get in the party.

It's a non-issue, but it was bothering me when I woke up this morning :cold_sweat:.


anthonyronda (commented):
In testing, I found that dropping a folder of Actors also includes non-character Actors

anthonyronda:
looks good!

PR #44, opened 2022-01-18
fix: individual initiative dice spam

RabidOwlbear:
Made changes to the individualInitiative function in the combat.js module:

- reformatted the chat message data to be more consistent with the preferred implementation for dice so nice.
- added a check to disable dice animation if the number of combatants exceeds 15.


anthonyronda:
@all-contributors please add @RabidOwlbear as a code contributor

PR #68, opened 2022-01-29
Fix #67 dice equals

### LINKED ISSUE #67: Ability checks with equals don't work
An ability check using the type equals (=) does not work correctly, if the roll matches the target number it will not signify a success.

engleback:
Fixes #67 ability dice rolls with type equals not working


anthonyronda:
LGTM

PR #91, opened 2022-02-15
feat: native support for Item Piles

anthonyronda:
Implements the Item Piles API to manually set the necessary values needed (specifically actorType and itemQuantityAttribute)

### DIFF
--- a/src/module/fvttModuleAPIs.js
+++ b/src/module/fvttModuleAPIs.js
@@ -4,8 +4,23 @@ export const registerFVTTModuleAPIs = () => {
     Hooks.once("item-piles-ready", async function () {
       if (ItemPiles.API.ACTOR_CLASS_TYPE !== "character")
         ItemPiles.API.setActorClassType("character");
-      if (!ItemPiles.API.ITEM_QUANTITY_ATTRIBUTE !== "data.quantity.value")
+      if (ItemPiles.API.ITEM_QUANTITY_ATTRIBUTE !== "data.quantity.value")
         ItemPiles.API.setItemQuantityAttribute("data.quantity.value");
+      if (
+        ItemPiles.API.ITEM_FILTERS !==
+        [
+          {
+            path: "type",
+            filters: "spell,ability",
+          },
+        ]
+      )
+        ItemPiles.API.setItemFilters([
+          {
+            path: "type",
+            filters: "spell,ability",
+          },
+        ]);
     });
   }
 };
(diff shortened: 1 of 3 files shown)


anthonyronda:
To test this PR: 

- Install the Item Piles module
- Enable the Item Piles module
- Go to Settings > Module Settings, scroll to Item Piles
- The actor type setting should say `character` and the item quantity attribute setting should say `data.quantity.value`

Known issues: 

Item Piles still shows the "this system is not compatible" dialog box

Willing to hear feedback: 

Do you like putting all of the "registering Module APIs/setup" into this file?

freohr:
First round of feedback

After installing Item Piles, and checking the configuration as you explained, I checked:

* Droping coins from an inventory to the ground creates a pile with the number of coins in it and removes them from my invenotyr :heavy_check_mark: 
* Droping items (e.g. weapons or armor) on the ground or on an existing pile remove the item from my inventory, but adds the item with a quantity of zero in the pile :negative_squared_cross_mark:

I can do more testing tomorrow and this weekend if needed :)

anthonyronda:
I've assessed the problem: we don't include a quantity field except for the generic `item` type (❗)

chrisesharp:
Are you aware of this? https://github.com/fantasycalendar/FoundryVTT-ItemPiles/issues/43

anthonyronda:
@chrisesharp yep! I've spoken with Wasp in the Fantasy Computerworks Discord server and after a back-and-forth, they're going to let us handle support via API and I've told them they're free to close the issue.

anthonyronda:
@freohr Should be ready to go now!

- there's been a template.json change. We probably don't have to write a compendium migration for this change, but keep an eye out for possible side effects.
- spells and abilities are now properly "filtered out"

freohr:
I've done a few more tests with your changes, so:
1. Dropping and Taking items works :+1: 
2. There's a warning when dropping Spells and Abilities so that's good :+1: , but when accepting the warning the ability dropped disappears (neither in the character sheet, nor in the pile itself, nor in the character sheet for the pile). 
    * I don't know if it should be changed, fixed, and if it's on us or something in the Item Piles modules.

Either way, looks good to me :slightly_smiling_face:

freohr (approved):

Haxxer:
Item Piles now fully supports systems that directly set its settings through the API, and suppresses the system incompatibility warning.

PR #93, opened 2022-02-17
Display containers as categories for ease of use + Inventory HTML/JS refactor

freohr:
**New Features**

* Content of containers is displayed as a category (like Weapons, Armor or Treasure) linked to the container item display, to enable ease of use and readability of their content
* Monsters' attack pattern groups are now separated by a small margin for ease of readability (a happy little accident I found neat while refactoring the internal structure of this display)

**Internal refactor**
Oh god so much stuff (<- this is not mandatory to include in the release notes)

* Item summary display (for weapons, armors, items, containers, treasure, monster abilities, spells, etc.) is no longer generated with JS, but is contained in an Handlebars template
* Item tags are now accessible in two data fields: `data.autoTags` and `data.manualTags`, which respectively represents tags derived from the item qualities (Armor weight, Weapon usage, Class needed for a Spell, Ability roll, etc.) and tags added manually to weapons (the system could be extended to other Items, e.g. to mark armors as magic, or treasure as cursed). Template usage have been adjusted accordingly. 
* Item categories internal structure is simpler (alternating Title -> Items -> Title -> Items -> etc.) instead of the monstrosity of divs inside divs inside divs (for the Inventory tab and the Spells tabs)
  * Same deal for Item element in the inventory lists, with a consistent structure across all kind of Items displayed
  * Which also means that the CSS has been adjusted to match the new structure
* Any event handling methods that was in either character-sheet.js or monster-sheet.js (or most of the time, in both, but with very subtle differences, like naming) has been factorized in actor-sheet.js, from which both JS classes inherits
* Speaking of event handling, methods bound to events and longer than 3 lines have been extracted as private methods (well, in methods prefixed with `_`, if this is how private methods are handled in JS) for code readability

### DIFF
--- a/src/module/actor/actor-sheet.js
+++ b/src/module/actor/actor-sheet.js
@@ -31,25 +31,15 @@ export class OseActorSheet extends ActorSheet {
     super.activateEditor(name, options, initialContent);
   }
 
-  _onItemSummary(event) {
+  _toggleItemSummary(event) {
     event.preventDefault();
-    let li = $(event.currentTarget).parents(".item"),
-      item = this.actor.items.get(li.data("item-id")),
-      description = TextEditor.enrichHTML(item.data.data.description);
+    const summary = $(event.currentTarget).closest(".item-header").next(".item-summary");
 
-    // Toggle summary
-    if (li.hasClass("expanded")) {
-      let summary = li.parents(".item-entry").children(".item-summary");
-      summary.slideUp(200, () => summary.remove());
+    if (summary.css("display") === "none") {
+      summary.slideDown(200);
     } else {
-      // Add item tags
-      let div = $(
-        `<div class="item-summary"><ol class="tag-list">${item.getTags()}</ol><div>${description}</div></div>`
-      );
-      li.parents(".item-entry").append(div.hide());
-      div.slideDown(200);
+      summary.slideUp(200);
     }
-    li.toggleClass("expanded");
   }
 
   async _onSpellChange(event) {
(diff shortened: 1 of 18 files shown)


anthonyronda:
I'm encountering some unexpected behavior with the new Containers feature

1. Begin with a character sheet that doesn't have any containers
2. Click Add from the Containers Header *twice.* Two New Containers appear under the Containers header.
3. Rename one New Container Backpack, and the other Ratpack.
4. Drag Ratpack into Backpack
5. Click Delete next to Ratpack.

Expected behavior: Ratpack is removed.
Actual behavior: Ratpack is not removed, console error.

```console
actor-sheet.js:87 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'reduce')
[Detected 1 package: system:ose]
    at OseActorSheetCharacter._removeItemFromActor (actor-sheet.js:87)
    at HTMLAnchorElement.<anonymous> (actor-sheet.js:385)
    at HTMLAnchorElement.dispatch (jquery.min.js:2)
    at HTMLAnchorElement.v.handle (jquery.min.js:2)
```

It was not previously possible to remove items until they were removed from a container, so this is new behavior.

anthonyronda:
In #91 I made a change that now initializes containers with an empty containerId like all other "physical" items. I am assuming this is a non-breaking change for this PR based on my own reading. What do you think?

My own testing came out great, so if you can give the thumbs up on the above we'll get this merged :smile:

anthonyronda (approved):
LGTM, thanks so much for all your help this weekend ❤️

src/module/actor/actor-sheet.js
```diff
@@ -190,6 +197,82 @@ export class OseActorSheet extends ActorSheet {
...
+    }
+  }
+
+  async _updateItemQuantity(event) {
+    event.preventDefault();
+    const item = this._getItemFromActor(event);
+
+    if (event.target.dataset.field === "value") {
+      return item.update({
+        "data.counter.value": parseInt(event.target.value),
```
anthonyronda, line 267: oops, just found this. Copied from somewhere? needs to be data.quantity.value

anthonyronda: Nvm we'll chat in Discord

PR #95, opened 2022-02-18
Lookup folder recursively to add characters to party

freohr:
Just basic stuff, I didn't check how folders contained each others when refactoring the party sheet, so there it is.

As a bonus, I added checks to avoid any computation when trying to drop non-actor folders and items on the party-sheet.


anthonyronda (approved):
LGTM

PR #96, opened 2022-02-18
Update config to prepare support for "Let me roll that for you!"

freohr:
Updated the config.js file to add the arrays needed by [Let me roll that for you!](https://github.com/League-of-Foundry-Developers/fvtt-module-lmrtfy).

Notes:
* I'll create the pull request on their side when this is integrated, I don't want to have users try that module and have it not work.
  * If anyone wants to test, you can checkout my forked branch [here](https://github.com/freohr/fvtt-module-lmrtfy/tree/feature/add-ose-support) 
* I don't know how to integrate roll of abilities (such as Thief skills), as they are handled as an collection of Items (with the roll logic actually in the Item, not the Actor it's attached to) in this system, and LMRTFY looks for methods on the actors to enable asking for rolls.

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -273,6 +273,8 @@ export class OseActor extends Actor {
   }
 
   rollCheck(score, options = {}) {
+    if (this.data.type !== "character") return;
+
     const label = game.i18n.localize(`OSE.scores.${score}.long`);
     const rollParts = ["1d20"];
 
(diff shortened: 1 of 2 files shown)


anthonyronda:
I really can't give much time to testing with your LMRTFY fork, but I'm satisfied that this doesn't break anything and if there's a LMRTFY error we'll patch it in later 🙂

PR #109, opened 2022-02-21
fix: reverting some dependency changes

anthonyronda:
Freohr and I found a bug in a dependency. It appears to be caused by inconsistent behavior with the preinstall script for forcing resolutions on dependencies. Reverting back to the old behavior


anthonyronda:
Sneaking in a fix relating to a change in `ose.js` location now that distribution js is rolled up into a single file

Issue #80, opened 2022-02-04
Errors when creating treasure tables


RabidOwlbear:
getting errors when trying to create treasure tables.

When a new table is created, and the treasure chest icon is clicked, if you try to add a result option it throws an error:
"Invalid tableResult range which must which must bee a length-2 array of ascending integers."

foundry.js:3856 Error: [6VgXAmoZOylmcLqY] Invalid TableResult range which must be a length-2 array of ascending integers
    at TableResultData._validateField (/C:/Users/redacted/Desktop/FVTT/FoundryVTT/Foundry%20Virtual%20Tabletop/resources/app/common/abstract/data.mjs:384)
    at TableResultData.validate (/C:/Users/redacted/Desktop/FVTT/FoundryVTT/Foundry%20Virtual%20Tabletop/resources/app/common/abstract/data.mjs:275)
    at TableResultData.update (/C:/Users//Desktop/FVTT/FoundryVTT/Foundry%20Virtual%20Tabletop/resources/app/common/abstract/data.mjs:454)
    at RollTableData.updateCollection (/C:/Users/redacted/Desktop/FVTT/FoundryVTT/Foundry%20Virtual%20Tabletop/resources/app/common/abstract/data.mjs:479)
    at RollTableData.update (/C:/Users/redacted/Desktop/FVTT/FoundryVTT/Foundry%20Virtual%20Tabletop/resources/app/common/abstract/data.mjs:431)
    at ServerDatabaseBackend._updateDocuments (/C:/Users/redacted/Desktop/FVTT/FoundryVTT/Foundry%20Virtual%20Tabletop/resources/app/dist/database/backend/server-backend.mjs:1)
    at async Semaphore._try (/C:/Users/redacted/Desktop/FVTT/FoundryVTT/Foundry%20Virtual%20Tabletop/resources/app/common/utils/semaphore.mjs:99)
_onSubmit @ foundry.js:3856

anthonyronda:
It seems as though a workaround is to compile the treasure table as a normal rollTable, write your Chances in the Weight column, and then click the Treasure button at the end. For me this method works without error (and indeed this may be the intended usage).

PR #112, opened 2022-02-21
Fix error on Treasure Table "update" and "add" + visual fixes

freohr:
Quick fix for the "Treasure Table" problems encountered in #80

### DIFF
--- a/src/module/treasure.js
+++ b/src/module/treasure.js
@@ -1,6 +1,6 @@
 export const augmentTable = (table, html, data) => {
   // Treasure Toggle
-  const isTreasureTable = table.object.getFlag("ose", "treasure");
+  const isTreasureTable = Boolean(table.object.getFlag("ose", "treasure"));
 
   let treasureTableToggle = $("<div class='toggle-treasure' title='Toggle Treasure Table'></div>");
   treasureTableToggle.toggleClass("active", isTreasureTable);
@@ -9,7 +9,7 @@ export const augmentTable = (table, html, data) => {
   head.append(treasureTableToggle);
 
   html.find(".toggle-treasure").click((ev) => {
-    const isTreasure = table.object.getFlag("ose", "treasure");
+    const isTreasure = Boolean(table.object.getFlag("ose", "treasure"));
     table.object.setFlag("ose", "treasure", !isTreasure);
   });
 
--- a/src/module/treasure.js
+++ b/src/module/treasure.js
@@ -1,31 +1,40 @@
 export const augmentTable = (table, html, data) => {
   // Treasure Toggle
+  const isTreasureTable = table.object.getFlag("ose", "treasure");
+
+  let treasureTableToggle = $("<div class='toggle-treasure' title='Toggle Treasure Table'></div>");
+  treasureTableToggle.toggleClass("active", isTreasureTable);
+
   let head = html.find(".sheet-header");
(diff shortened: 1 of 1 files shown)


anthonyronda:
This doesn't fix it for me, I'm afraid. I'm experiencing the same behavior in this pull request branch as on main branch, the error described in issue 80.

In particular, I can confirm that the validation function that's causing the error is blocking the table entry from writing to the table *before* any of the code committed here is run. If you're experiencing otherwise, there may be a race condition.

anthonyronda (changes requested):
Does not work on my machine, error #80 persists

anthonyronda:
As a follow up to this, we discovered some problems with a test build based on main branch that had not surfaced on my dev builds. I'll revisit this PR after we've gotten that resolved

src/module/treasure.js
```diff
@@ -1,32 +1,41 @@
 export const augmentTable = (table, html, data) => {
   // Treasure Toggle
+  const isTreasureTable = table.object.getFlag("ose", "treasure");
```
anthonyronda, line 3: getFlag() returns undefined for a table where the flag has not yet been created by the click event

src/module/treasure.js
```diff
@@ -1,32 +1,41 @@
...
+
+  let treasureTableToggle = $("<div class='toggle-treasure' title='Toggle Treasure Table'></div>");
+  treasureTableToggle.toggleClass("active", isTreasureTable);
```
anthonyronda, line 6: [.toggleClass( className, state )](https://api.jquery.com/toggleClass/#toggleClass-className-state)

    className
    Type: [String](http://api.jquery.com/Types/#String)
    One or more classes (separated by spaces) to be toggled for each element in the matched set.
    state
    Type: [Boolean](http://api.jquery.com/Types/#Boolean)
    A boolean (not just truthy/falsy) value to determine whether the class should be added or removed.

While we're giving undefined for the state, it appears as though toggleClass treats it as true with no error returned. This results in unexpected behavior: it appears to default to toggled as active even though there's no flag by default.

anthonyronda (changes requested):
Sorry I found another nitpick!

The toggle button is defaulting to style class `active` on tables without a `treasure` flag. I think the expected behavior is to not show as `active` by default, while still not adding the `treasure` flag unless the treasure button is clicked.

I left comments to show why this is happening.

src/module/treasure.js
freohr: Good point, I didn't think of that either (still need to get used to the joy of `undefined` coming from C++). Would `getFlag() || false` enforce a boolean value?
This should be the only place where we change this flag for this Table object, but I don't know what would happen if something else put a flag that wasn't a boolean value on this object with this name.

freohr: I just thought of it, but is it an accepted practice if I use `!!getFlag()` to force a boolean value?

anthonyronda: I think you can Boolean(getFlag()) too. I don't mind either way

src/module/treasure.js
freohr: I understand, I intended to be a bit smarter than a ternary with 2 raw strings, and take advantage of JQuery for clarity. I put my questions regarding the handling of the `active` var above.

anthonyronda (approved):
Works for me now

Lets Get This Merged!

Issue #115, opened 2022-02-23
[1.3.7] Weapon attacks can't be rolled by clicking the weapon icon


anthonyronda:
Reported by DeanL on the Unofficial OSE Discord server

This is a showstopping bug for 1.3.7. 

Please uninstall and reinstall OSE to rollback to 1.3.6. Your world and compendiums will not be affected by this rollback.

anthonyronda:
If you'd really rather not rollback to 1.3.6, you can drag weapons to the macro bar to roll them from there.

PR #118, opened 2022-02-23
Fix rolling attacks with weapons from inventory and from containers

freohr:
I missed the needed class in the Item HTML header to enable the roll event, and I also added it for items inside containers


anthonyronda:
Thanks for the quick turnaround on this! Tracked in #115

anthonyronda (approved):
LGTM

PR #119, opened 2022-02-23
Fix/display tags item directory

freohr:
I moved the tag list block, displaying information in the item summary, to a template, and used that template to fix the display of tags in the sidebar directories


anthonyronda (approved):
LGTM

PR #120, opened 2022-02-23
Re-add a HTML class required for abilities drag&drop from inventory

freohr:
I removed a class on the HTML of the Item list elements, but the default drag&drop Foundry implementation depends on it, so I added it back


anthonyronda (approved):
Alright almost release time!

PR #128, opened 2022-02-28
actor drop fix

RabidOwlbear:
This seems to fix the issue of actors not being able to be added to the party sheet.


anthonyronda (approved):
LGTM, thanks for fixing this as quickly as you did!

Issue #85, opened 2022-02-13
OSE System + Maestro not working


odhinn33:
OSE System worked perfectly with Maestro once. Not anymore. For example with item tracks: not sure about the problem but many new systems needs to add the item-id in the chat messages that are created from item rolls. Then Maestro module should be able to work with them.

anthonyronda:
Thanks for reporting this, first I've heard this issue. I don't use it myself but I'd like to fix this incompatibility. I have a feeling I know how it happened, but do you know which version of OSE broke compatibility?

odhinn33:
> Thanks for reporting this, first I've heard this issue. I don't use it myself but I'd like to fix this incompatibility. I have a feeling I know how it happened, but do you know which version of OSE broke compatibility?

Unfortunately not, I don't remember. Thank you for your attention.

anthonyronda:
Twas a very easy bug to fix. It's something that got missed in the V9 compatibility update (Maestro should still work in 0.8.x too).

PR #130, opened 2022-02-28
fix: missing underscore broke Maestro

anthonyronda:
Now successfully passing actor and item ids with the weapon attack roll chat card template


anthonyronda:
Fixes #85

PR #135, opened 2022-03-03
Armor tag

RabidOwlbear:
added tagging to 'item' and 'armor' Item types.
fixed overflow issue with character sheet inventory tab item tag display.


anthonyronda:
LGTM

Issue #83, opened 2022-02-10
Feature Request: Advanced Fantasy Optional Rules


arcos:
As the title suggests, this is a feature request to add the advanced fantasy optional mechanics.

Including:

- separating races and classes (new place on the character sheet for race)
- multiclassing (so separated slots for experience, fractional hp, etc.)
- secondary skills (this might already be doable via the abilities tab)
- weapon proficiency (this might already be doable via the abilities tab)

crazyclausewitz:
The Advanced Fantasy Rules are not covered by the Open Game License. So i guess the reason there is no advanced fantasy stuff is mainly because of copyright issues.

arcos:
This would just add the ability to use the rules, not the content. The content is the copyright issue, the mechanics are not covered by that.

lonpalmer:
Is this feature on the back burner?  Any way we can vote to pull this forward?

Issue #97, opened 2022-02-19
Character sheet redesign


anthonyronda:
Godforsken, a member of the Unofficial OSE on Foundry VTT community, has created a new design for the character sheet.

[View the design in Figma](https://www.figma.com/file/p3Nir2WTXVgKsMniryYgqR/OSE-Character-sheet)

We are still developing a process around accepting open source design contributions and would love to hear from more designers! This is the first public step of design review, although we've had multiple rounds of generating feedback and requesting changes within the contributor community already (you can see much of that in the Figma comments).

We welcome additional comments in this GitHub Issue and in Figma (I remind you that the VTT Red [Code of Conduct](https://github.com/vttred/.github/blob/main/CODE_OF_CONDUCT.md) applies to your Figma comments).

anthonyronda:
@all-contributors please add @Godforsaken84 for design

PR #136, opened 2022-03-04
Adds multiclassing and races to the character sheet

arcos:
I asked for this [feature](https://github.com/vttred/ose/issues/83) and got impatient (and amazingly! had some time to hack at it). 
So I implemented races and multiclassing, which I think are the main differences between Classic Fantasy and Advanced Fantasy.

image


anthonyronda:
Hi Andrea, thanks for your pull request! I’d love to accept it right away, but there will be a delay due to having to check with Necrotic Gnome as to whether non-SRD character sheet features abide with their trademark license. We’ve already been given one special exception, and I’d rather not take too much advantage of this generosity.

anthonyronda:
There may also be design considerations (see #97) and party sheet considerations (users expecting this to work with the XP-granting feature of the party sheet)

arcos:
Necrotic Gnome should be fine with this, but I think you are right we should give Gavin a chance to respond. We aren't reproducing any of his content, just implementing rules that he himself repackaged from WOTC/TSR. Let me look at https://github.com/vttred/ose/issues/97 and the party sheet to improve on what I've done so far.

arcos:
@anthonyronda I added the party overview xp dispersal support, the design I guess depends on the new code to support that design. Is that an open PR?

anthonyronda:
The design is a branch but the implementation is far from done!

If you're on Discord, the Unofficial OSE on Foundry VTT Discord server is where I'm communicating with the Designer outside of Figma comments (the thread is public). Join if you'd like! https://discord.gg/thab6fGV

mummson:
@arcos Couldn't this be done as a separate addon module for this system? wouldn't that solve the problem if Necrotic Gnome wouldn't want this as a feature to the game system?

IGALEGOI:
Would it be interesting to add fields like HP and HD separately for each class? That allows you to write down the HP per class, as well set the hit dice per class, to roll the HD according to the amount of classes: HD/2 or HD/3.

anthonyronda:
We're still treating this PR as "pending a decision from Necrotic Gnome."

Gavin is aware we'd like to merge it while keeping the Old-School Essentials trademark exception they've given us. He doesn't need any prodding about this request specifically, though telling NG in general that you'd like official stuff on Foundry is ok.

SamJulson:
Hi! Hate to revive an issue but I was wondering if this has had an update, considering I was also ready to implement this feature. Thanks in advance!

PR #145, opened 2022-03-16
fix: settings not persisting on save changes

RabidOwlbear:
bug: system settings are not being correctly saved.
fix: removed the onChange key from the system settings objects.


freohr (approved):

anthonyronda:
LGTM

PR #146, opened 2022-03-16
Add hook to show party sheet from macros

freohr:
I wanted to have a way to show the party sheet from a macro to design a landing page for my campaign, so I added a way to display the Party Sheet with a hook call, which can be called from a macro, using the way "Forien's Quest Log" does it as an example.

Also, this fixes a bug where you get multiple party windows when you click on the party button multiple times

### DIFF
--- a/src/module/party/party-sheet.js
+++ b/src/module/party/party-sheet.js
@@ -103,7 +103,7 @@ export class OsePartySheet extends FormApplication {
       return;
     }
 
-    const actors = this.object.documents;
+    const actors = game.actors;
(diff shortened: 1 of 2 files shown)


anthonyronda:
The party sheet drag drop behavior is broken in this patch.

I'm afraid I don't have time to diagnose. It will have to wait until next release

anthonyronda (changes requested):
Drag Drop behavior is broken in this patch

anthonyronda:
LGTM

PR #148, opened 2022-03-17
Fix: tweaks icon not showing for non gm users

RabidOwlbear:
Fixed incorrect property name that was preventing the tweaks menu from being accessible on a user owned actor sheet.


anthonyronda:
Nice catch, but *should* players have access to the Tweaks menu? What is the use case?

I'll accept this PR since this seems to be the intended behavior, but my guess is that at least one GM will ask us to change it back.

anthonyronda (approved):
LGTM

PR #165, opened 2022-04-15
Fix the display of the resulting chat card when rolling treasure tables recursively

freohr:
(Edit: removed because I reworded the title)


anthonyronda (approved):
LGTM

PR #175, opened 2022-05-05
Update pt-BR.json

igorteuri:
- Fix some terms to the official translation in portuguese
- Change Salvaguarda to Jogada de Proteção
- Change Magia for Feitiços
- Fix some acronyms
- Fix the order of some terms to make more sense in portuguese


igorteuri:
Thank you a lot @anthonyronda for your work with this project!

PR #180, opened 2022-05-14
Individual initiative fix

RabidOwlbear:
This fixes an error thrown due to a bad path to the template used for individual initiative rolls. 
Error was caused by an extraneous 'dist/' in the file path.


anthonyronda:
Fug that one's on me, thanks for catching it

anthonyronda (approved):
LGTM

PR #183, opened 2022-05-21
Ability chat card render fix

RabidOwlbear:
Fixes the ability chat card rendering bug by removing an extraneous /dist in the path.
removed a couple console logs and empty lines


src/module/combat.js
```diff
@@ -14,6 +14,7 @@ export class OseCombat {
     };
   }
   static async rollInitiative(combat, data) {
+    console.log('roll init')
```
anthonyronda, line 17: oop, I'll remove this

anthonyronda (approved):
The review passed, but I'll close with changes, and merge #184 instead

PR #184, opened 2022-05-23
Fix/ability chat card render

anthonyronda:
#183 minus the console logging


anthonyronda:
Thanks @RabidOwlbear for the fix!

Issue #193, opened 2022-06-09
Shopping cart numbers off


aramyle:
Any items with a quantity (rations, rope, etc.) are multiplied by the quantity for the price. Rations costs 105gp instead of the 15gp it should because a ration comes with 7/7.

anthonyronda:
Thanks for reporting that because I don't have an issue with that yet. Was reported in Discord, and it's indeed an unintended behavior of the intended implementation.

My current fix, provided I can confirm it covers all cases, is to only multiply the cost when there isn't a maximum quantity for the item.

Issue #190, opened 2022-06-03
Improves & Translation on the new Shopping cart


WallaceMcGregor:
I've made a few tests with the amazing new shopping cart and would be great adding some little changes.

The first one It would be to be able of using translated coins. It only works with GP. Also, If the added starting gold could be translated It will be great.

The second one is to show only the item name and the total cost in the chat card. Descriptions of items can use too much space.

![chrome_REi6p0Zixe](https://user-images.githubusercontent.com/17795541/171789496-f8ec564e-fd61-4ed3-aac5-e0017d03f974.jpg)

anthonyronda:
In a v2 of this feature, I'd like to make it useful for any new merchant visit. For now it's better to use OSR Item Shop for that

WallaceMcGregor:
> For now it's better to use **OSR Item Shop** for that

Is that a Foundry module? I didn't found It and I'm interested 😋

anthonyronda:
@WallaceMcGregor https://github.com/RabidOwlbear/osr-item-shop

anthonyronda:
The suggestion to use a localized abbreviation for GP has been implemented in #194. We will need to wait for new translations to be available.

I decided to wait on the dropdown item summary until some other time

PR #194, opened 2022-06-09
Fix GP cost multiplying when a max quantity is defined

anthonyronda:


anthonyronda:
Fixes #193 and also the translation half of #190

PR #198, opened 2022-07-05
Item summary fix

RabidOwlbear:
Fixes problem with character sheet item summaries not displaying correctly for items with a max quantity value.


anthonyronda:
Hey there, reviewing now

anthonyronda:
When I reviewed this two months ago, I tried to determine a method that keeps the animation (or at least determine why jQuery seemed to be causing this).

At this point I think removing our few bits of jQuery from the system is probably for the best anyway, so I'll accept as is.

Issue #222, opened 2022-09-06
Languages and dungeon skill roles bugged


anthonyronda:
### What happened?

1. Try to type anything into the Language or dungeon skill boxes, press enter
2. The box remains empty (Languages) or remain at 0 (dungeon skills)

### What is the expected behaviour?

The input is saved to the field when it's a valid input

### Foundry VTT Core Version

V10

### Old-School Essentials Core Version

1.6.2

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #220, opened 2022-09-06
Rollable items bugged


anthonyronda:
### What happened?

1. Click a rollable item
2. It doesn't do anything, error logged to console (below)
![image](https://user-images.githubusercontent.com/1410433/188688615-2dfed067-f8a7-48a9-aed8-6c4bba714495.png)

### What is the expected behaviour?

The roll is added to the chat log

### Foundry VTT Core Version

V10

### Old-School Essentials Core Version

1.6.2

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #224, opened 2022-09-06
Fix languages, dungeon skills, and rollable items (abilities, weapons, spells) in V10

anthonyronda:


anthonyronda:
Fixes #222 and #220

PR #235, opened 2022-09-20
party sheet/container fix

RabidOwlbear:
fixes the container item duplication bug, and fixes the broken party sheet. Have not tested on a v9 client.


anthonyronda (approved):
LGTM

PR #236, opened 2022-09-21
item templates temp fix

RabidOwlbear:
fixes version compatibility for item sheets


anthonyronda (approved):
Let's accept this as-is

PR #239, opened 2022-09-26
Add a placeholder text to the Party overview when empty

freohr:
I noticed that the question "How do I add characters to the party?" is asked a bit too much in the Discord, so I added a placeholder text that should have been there when I reworked this window >.>


anthonyronda:
Very thoughtful PR. Putting this on the list to review today

anthonyronda (approved):
LGTM

PR #253, opened 2022-10-12
Replace gulp with rollup

maschill92:
The existing gulp build processed leveraged rollup to build JS and TS files.

Using the static file copy plugin `@web/rollup-plugin-copy`, the static files are copied using rollup.

Gulp also provided a command for creating and removing the dev symlink. That functionality has been retained with `tools/symlink.js`. Symlink commands have been added to `package.json` to retain that funcionality for devs.


rollup.config.js
```diff
@@ -1,14 +1,29 @@
...
+  input: "src/ose.js",
+  output: {
+    dir: "dist/",
+    format: "es",
+    sourcemap: true,
+  },
+  plugins: [
+    nodeResolve(),
+    typescript(),
+    scss({
```
wyrmisis, line 19: When used from `build:watch`, is this going to trigger a reload of the page for updates to our CSS?

I'd be a fan of getting it to hot reload instead, but I get if we can't.

rollup.config.js
```diff
@@ -1,14 +1,29 @@
...
-      format: "es",
-      sourcemap: true,
-    },
-    plugins: [nodeResolve(), typescript()],
-  });
+const staticFiles = ["lang", "packs", "templates"];
+
+export default defineConfig({
+  input: "src/ose.js",
+  output: {
```
wyrmisis, line 11: Can we minify our JS payload?

src/ose.js
```diff
@@ -16,6 +16,7 @@ import * as party from "./module/party";
 import { OseCombat } from "./module/combat";
 import * as renderList from "./module/renderList";
 import { OsePartySheet } from "./module/party/party-sheet";
+import "./ose.scss";
```
wyrmisis, line 19: What's this do if we're already pulling our styles in from `system.json`?

tools/symlink.js
```diff
@@ -0,0 +1,70 @@
...
+);
+
+const linkDirectory = path.resolve(
+  foundryConfig.dataPath,
+  "Data",
+  "systems",
+  foundryConfig.symLinkName
+);
+
+if (argv.clean) {
```
wyrmisis, line 32: ⛳ 
```js
handleProcess(argv.clean ? removeSymlink() : createSymlink());
```

tools/symlink.js
```diff
@@ -0,0 +1,70 @@
...
+  handleProcess(removeSymlink());
+} else {
+  handleProcess(createSymlink());
+}
+
+async function createSymlink() {
```
wyrmisis, line 38: This is async, but not awaiting anything. Should there be an await here?

src/ose.js
maschill92: This is included to convert `scss` to `css`. The rollup plugin for `scss` captures that bundle and emits it to `dist/` in the location that `system.json` expects it.

tools/symlink.js
anthonyronda: possibly a question for me, since this is mainly a copy/paste from the gulpfile. Checking what happened here now

tools/symlink.js
```diff
@@ -0,0 +1,70 @@
...
+  const systemsDirectory = path.resolve(linkDirectory, "..");
+  if (!fs.pathExistsSync(systemsDirectory)) {
+    throw new Error(
+      `Link directory parent folder "${systemsDirectory}" not found.`
+    );
+  }
+
+  if (!fs.existsSync(linkDirectory)) {
+    // link path doesn't exist, create it.
+    fs.symlinkSync(path.resolve(__dirname, ".."), linkDirectory);
```
anthonyronda, line 49: this was originally awaited. Perhaps it's not necessary given the way the functions are now split up? If it works as is, then yeah async isn't necessary and we can ditch it from the function declaration

src/ose.js
anthonyronda: seems appropriate, it fits the rollup-plugin-scss usage example
image

tools/symlink.js
maschill92: I made it async as the `handleProcess` function expects the pass parameter to be a Promise (the call to `.then`).

At the very least, I'll rename `handleProcess` to `handlePromise`. I'm pretty sure that's what I _wanted_ to name it in the first place, but mistyped.

maschill92: I could also return `Promise.resolve` or `Promise.reject`, but making it `async` was more straightforward in my mind.

maschill92: I'm realizing now, that I can just use the `*Sync` functions across the board in this file. I'll do that and remove the need for `handleProcess`.

tools/symlink.js
maschill92: I'll remove the need for `await` and `async` in the file all together by using the `*Sync` functions.

rollup.config.js
anthonyronda: I anticipate a conflict with "double sourcemapping" again if we do

https://stackoverflow.com/a/63235210

Perhaps one of you could confirm?

rollup.config.js
maschill92: Remove livereload for now

rollup.config.js
maschill92: `tsconfig.json` doesn't have `sourcemaps` compiler option defined. Sourcemaps are looking healthy

maschill92: Also, added `rollup-plugin-terser` for minification of JS and `outputStyle: "compressed"` for CSS

rollup.config.js
maschill92: Update pushed, we have in-place css reloading now

anthonyronda (approved):
LGTM

anthonyronda:
@all-contributors please add @maschill92 for code

PR #259, opened 2022-10-14
Fix Typescript build warnings

maschill92:
Most of the build warnings were type issues with calls to `games.settings.register`. Added type def override to `SystemData` that defines `game.system.id` as a const `"ose"`.


anthonyronda (approved):
LGTM

Issue #254, opened 2022-10-13
No Tag Icons


justinthejaguar:
### What happened?

Toggling icon tags in equipment does not make the tags appear as icon on the character sheet.

### What is the expected behaviour?

Toggling options should appear as icons on the character sheet.

### Relevant Errors and Warnings

```shell
N/A
```

### Foundry VTT Core Version

V10

### Old-School Essentials Core Version

1.6.7

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #260, opened 2022-10-16
Bug/#254-no-tag-icons

maschill92:
Fix for #254.

Correct access for item tags.

Correct access for all assets in development when system ID is 'ose-dev'

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -35,7 +35,7 @@ export default defineConfig([
     plugins: [
       nodeResolve(),
       typescript(),
-      terser(),
+      !isWatchMode && terser(),
       copy({
         patterns: staticFileFolders.map((folderName) => `${folderName}/**/*`),
         rootDir: "./src/",
--- a/src/module/helpers.ts
+++ b/src/module/helpers.ts
@@ -36,8 +36,12 @@ export const registerHelpers = async function () {
     return Math.round(parseFloat(weight) / 100) / 10;
   });
 
-  Handlebars.registerHelper("getTagIcon", function (tag) {
-    return CONFIG.OSE.tag_images[tag as InventoryItemTag];
+  Handlebars.registerHelper("getTagIcon", function (tagValue: string) {
+    let tagKey = (Object.keys(CONFIG.OSE.tags) as InventoryItemTag[])
+      // find key for the tag display name who's name matches the provided tag text.
+      .find((findTagName) => CONFIG.OSE.tags[findTagName] === tagValue);
+    // if that tag key is found, return the image for the tag key
+    return tagKey ? CONFIG.OSE.tag_images[tagKey] : null;
   });
 
   Handlebars.registerHelper("counter", function (status, value, max) {
--- a/src/module/helpers.ts
+++ b/src/module/helpers.ts
@@ -60,6 +60,10 @@ export const registerHelpers = async function () {
     return `${OSE.systemPath()}${relativePath}`;
   });
 
+  Handlebars.registerHelper("asset", function (relativePath) {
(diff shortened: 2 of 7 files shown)


src/module/helpers.ts
```diff
@@ -36,8 +36,12 @@ export const registerHelpers = async function () {
     return Math.round(parseFloat(weight) / 100) / 10;
   });
 
-  Handlebars.registerHelper("getTagIcon", function (tag) {
-    return CONFIG.OSE.tag_images[tag as InventoryItemTag];
+  Handlebars.registerHelper("getTagIcon", function (tagValue: string) {
```
maschill92, line 39: Changes to this helper are the "actual" fix for the bug.

The rest is to get all paths to `/system/<system_id>/assets` to display correctly during development

src/ose.js
```diff
@@ -134,7 +134,7 @@ Hooks.on("renderSidebarTab", async (object, html) => {
     const styling =
       "border:none;margin-right:2px;vertical-align:middle;margin-bottom:5px";
     $(
-      `<button data-action="userguide"><img src='systems/ose/assets/dragon.png' width='16' height='16' style='${styling}'/>Old School Guide</button>`
+      `<button data-action="userguide"><img src='${OSE.assetsPath}/dragon.png' width='16' height='16' style='${styling}'/>Old School Guide</button>`
```
wyrmisis, line 137: While you're here, could you add `type="button"` to this button? While not explicitly necessary (an unset button type is equivalent to a button of type "submit" and I don't think this button is in a form), I think it'll be good to model good behavior for when we start into the new sheets.

rollup.config.js
```diff
@@ -35,7 +35,7 @@ export default defineConfig([
     plugins: [
       nodeResolve(),
       typescript(),
-      terser(),
+      !isWatchMode && terser(),
```
wyrmisis, line 38: heck yeah

src/templates/chat/roll-treasure.html
```diff
@@ -1,26 +1,26 @@
...
-    </ol>
-  </div>
+<div class="ose chat-card treasure-card">
+  <header class="card-header flexrow">
+    <img src="{{asset '/treasure.png'}}" title="{{table.name}}" width="36" height="36" />
+    <h3>{{table.name}}</h3>
+  </header>
+  <div class="card-content">
+    <ol class="treasure-list">
+      {{#each treasure as |t|}}
```
wyrmisis, line 8: While we're in here... thoughts on adding an `#else` block to this `#each` for when there's no treasure, to let the user know they didn't get any treasure?

wyrmisis (approved):
I'm good for what's here, just some minor "while you're in there" comments.

wyrmisis:
Oh hey don't forget to connect this to the issue this PR fixes!

src/templates/chat/roll-treasure.html
maschill92: @anthonyronda what's the general approach to getting adding new localization strings?

anthonyronda: Short answer is "add English strings as commits, omit all other languages"
Language contributors will get an email if they have email alerts set whenever https://crowdin.com/ autosync discovers new strings

I will add this to my list for improving documentation

src/ose.js
anthonyronda: I see that this has been added

src/templates/chat/roll-treasure.html
anthonyronda: noting that "No Items" was added

anthonyronda (approved):
LGTM

PR #269, opened 2022-12-01
fix: hit die roll adds con modifier for each level

### LINKED ISSUE #195: [Bug] HD roll does not calculate CON modifier correctly
On a PC sheet, when you click the HD button, the CON modifier is applied only once instead of once per hit dice (max 9 times). 

Thank you!

anthonyronda:
Fixes #195

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -337,7 +337,7 @@ export class OseActor extends Actor {
     const label = game.i18n.localize(`OSE.roll.hd`);
     const rollParts = [actorData.hp.hd];
     if (actorType == "character") {
-      rollParts.push(actorData.scores.con.mod);
+      rollParts.push(actorData.scores.con.mod*actorData.details.level);
     }
 
     const data = {


src/module/actor/entity.js
```diff
@@ -337,7 +337,7 @@ export class OseActor extends Actor {
     const label = game.i18n.localize(`OSE.roll.hd`);
     const rollParts = [actorData.hp.hd];
     if (actorType == "character") {
-      rollParts.push(actorData.scores.con.mod);
+      rollParts.push(actorData.scores.con.mod*actorData.details.level);
```
wyrmisis, line 340: Do we care about how this relates to OSE AF's multiclassing rules?

anthonyronda: I will add this to #270, which I just wrote up. For now I don't foresee being able to do this without some kind of enhancement to the leveling experience (and we're treating this as a bugfix)

wyrmisis (approved):

PR #272, opened 2022-12-01
fix: async restores treasure result text

### LINKED ISSUE #268: Treasure Tables Inline Dice Rolls are displaying [object Promise] in chat
### What happened?

When rolling the treasure for monsters using the treasure tables provided, the chat window displays [object Promise]

When rolled, treasure should resolve the inline roll and display the text properly.  For example, if the treasure result is: 
[[1d4 * 1000]] sp.
Then the resulting text should be something like:
2000 sp.
Screenshot 2022-11-30 at 12 16 39 PM

_No response_

_No response_

### Foundry VTT Core Version

V10

1.6.8

Chrome

### Code of Conduct

anthonyronda:
Fixes #268 

This was a case of the default enriched text behavior switching to async, and we missed this particular usage of it

### DIFF
--- a/src/module/treasure.js
+++ b/src/module/treasure.js
@@ -51,21 +51,20 @@ export const augmentTable = (table, html, data) => {
   });
 };
 
-async function percentageRoll (chance) {
-  const roll = new Roll("1d100");
-  roll.evaluate({ async: true });
-  return roll.total <= chance;
-};
-
-async function drawTreasure(table, data) {
+function drawTreasure(table, data) {
+  const percent = (chance) => {
+    const roll = new Roll("1d100");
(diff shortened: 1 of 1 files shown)


wyrmisis:
I did some tinkering with this, and found that, with the enrichment stuff being async by default now, we had a bunch of race conditions.

We might want to consider giving this a bit of a closer look, maybe consider rewriting some of it?

anthonyronda:
Ultimately you're correct about this, and I just don't think that's what I want holding back 1.7.0

I've confirmed a minimal fix that uses the `async: false` flag on the enrichHTML functions and makes no other changes. We will of course be looking for an async solution before V12 (ideally before V11)

PR #274, opened 2022-12-09
fix: Party sheet ergonomic fixes

wyrmisis:
## Intent
This changeset allows the user to drag a folder of characters into the party sheet. It also allows the user to drag an actor from the party sheet to the scene.


anthonyronda:
adding @freohr as reviewer who has previously been interested as kind of a "code owner" of the party sheet. @freohr feel free to ask questions about these changes

anthonyronda:
if we don't hear back for a few days I think we're safe to merge

freohr:
LGTM as well, doing it via the UUID to get the exact data is cleaner than what I had done :+1:

freohr (approved):

PR #275, opened 2022-12-13
feat: Add a data model for Monster actors

wyrmisis:
## Intent

This changeset adds a data model for the Monster actor type. it also scrubs some now-obsolete code from sheets and entities.


anthonyronda (changes requested):
With inventory enabled, monsters have a bugged encumbrance bar

image

So I think my only request for change is an "if character" handlebar block around the encumbrance bar in the inventory sheet template

anthonyronda (approved):
LGTM

PR #279, opened 2022-12-30
fix: Monster Data Model bugfixes

wyrmisis:
## Intent

This changeset fixes some stuff we didn't catch when monster data models went in the first time. Oops!


anthonyronda (approved):
LGTM :)

PR #289, opened 2023-01-11
fix: Additional v1.7.x fixes

wyrmisis:
## Intent

This changeset fixes some additional bugs found by users, namely:
* The character generator once again generates stats
* Treasure items created from the character sheet once again go into the correct heading
* Items of `item` type should once again display their descriptions

Note: I'm not 100% sure how to test the item description fix, but I'm happy to take suggestions!

### DIFF
--- a/src/module/actor/actor-sheet.js
+++ b/src/module/actor/actor-sheet.js
@@ -371,7 +371,7 @@ export class OseActorSheet extends ActorSheet {
   _createItem(event) {
     event.preventDefault();
     const header = event.currentTarget;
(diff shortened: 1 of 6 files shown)


anthonyronda (changes requested):
`item` fix still elusive.

Check for bug by comparing `item/item-sheet.js`'s `data.enriched` in `getData()` with each item type's template file. `armor` and `item` templates say `system.enrichedDescription` instead of `enriched.description`

anthonyronda (approved):
LGTM

PR #292, opened 2023-01-13
Updates to Item Piles system integration

Haxxer:


Haxxer:
It utilizes the new Item Piles API method `addSystemIntegration` which allows us to increment the `VERSION` property when it is changed, which will gracefully update users to newer versions of data structures.

Haxxer:
This last commit ensures that items of type `weapon`, `armor`, and `container` are always considered to be separate documents and cannot stack, and makes sure that items of type `item` with the `treasure` property enabled will be considered as a separate document to items with it disabled.

anthonyronda (approved):
LGTM

Issue #124, opened 2022-02-24
[Enhancement] A new setting to toggle Ctrl/Cmd+click behavior when rolling attacks


woodwardj:
If I need to roll multiple NPC attacks, it is nice to just click x number of attacks instead of clicking the attack and then the pop-up for each attack.   It seems that the default should be no situational modifiers, so that should require the least number of clicks.  DCC has a nice setup where you Ctrl-Click to modify a dice roll, or it just rolls with the default roll settings.

woodwardj:
I just realized that Ctrl-click can be used to quickly roll without the modifier popup dialog.   It would be nice to have the option to invert the behavior but it is completely workable as is.

anthonyronda:
Speaking of Ctrl-click, it appears to be broken for me on Mac 😮. Neither Ctl+click nor Cmd+click do anything

I believe we aren't using the V9 keybind API which would fix that

praetor:
Having a modification dialog option for other non-attack rolls would be nice. For abilities which are clicked to roll, being able to add situational modifiers would be a great enhancement.

anthonyronda:
We've just had a request on the OSE Community Discord for the same feature. 

Changing the issue title to help with discovery

### Current behavior 

When you click on an attack roll, there's a modifier dialog box that asks to input a roll input before rolling. You may skip this by holding ctl/cmd as you click. There is not a setting to invert, showing the prompt only while holding ctl/cmd. There is not a modifier window for every type of item roll.

### Requested behavior

There is a new setting toggle the user can select. It will allow users to make all attack rolls without showing the modifier dialog box.

PR #294, opened 2023-01-14
[enhancement] Added option to revert Ctrl/Meta behavior

bakbakbakbakbak:
resolves vttred/ose#124

Fixed the Mac-issues with Meta key for rolling, and added an option to reverse the behavior of clicking on a roll with ctrl/meta held down.

### DIFF
--- a/src/e2e/actor/character.e2e.test.js
+++ b/src/e2e/actor/character.e2e.test.js
@@ -5,7 +5,19 @@ export const options = {
   displayName: 'The Character Sheet'
 };
 
-export default ({before, beforeEach, after, describe, it, expect, ...context}) => {
+/**
+ * If there are dialogs, close them.
+ * @returns {Promise} the promise from closing dialogs
+ */
+export const closeRollDialog = async () => {
+  const openDialogs = Object.values(ui.windows).filter(o => o.options.classes.indexOf('dialog') > -1);
+
+  openDialogs?.forEach(async (o) => {
+    await o.close();
+  }) 
+}
+
+export default ({before, beforeEach, after, afterEach, describe, it, expect, ...context}) => {
   const testCharacterName = 'Quench Test Character';
   const prepareActor = async (data) => {
     await trashChat();
@@ -17,11 +29,26 @@ export default ({before, beforeEach, after, describe, it, expect, ...context}) =
       type: 'character'
     });
   };
+
   const testActor = () => game.actors.getName(testCharacterName);
   const trashActor = () => testActor()?.delete();
   
   const rollNoMods = async (key, rollFn) => {
     await testActor()[rollFn](key, {fastForward: true});
+    await waitForInput();
+    expect(game.messages.size).to.equal(1);
+  }
+
+  const rollNoModsSkipDialog = async (key, rollFn) => {
+    const ctrl_down = new KeyboardEvent('keydown', {ctrlKey: true});
+    await testActor()[rollFn](key, {event: ctrl_down});
+    await waitForInput();
+    expect(game.messages.size).to.equal(1);
+  }
+
+  const rollNoModsSkipDialogMeta = async (key, rollFn) => {
+    const meta_down = new KeyboardEvent('keydown', {metaKey: true});
+    await testActor()[rollFn](key, {event: meta_down});
     await waitForInput();
     expect(game.messages.size).to.equal(1);
   }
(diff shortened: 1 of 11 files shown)


src/module/actor/entity.js
```diff
@@ -134,7 +134,13 @@ export class OseActor extends Actor {
       details: game.i18n.format("OSE.roll.details.save", { save: label }),
     };
 
-    let skip = options?.event?.ctrlKey || options.fastForward;
+    const invertedCtrlBehavior = game.settings.get(game.system.id, "invertedCtrlBehavior");
```
wyrmisis, line 137: We're using this in a few places. We should extract it into its own method.

wyrmisis (changes requested):

src/module/actor/entity.js
bakbakbakbakbak: Indeed. I made an effort for it in the latest commit, also to realize there are no tests written for monster rolls.

wyrmisis: re: no tests for, well, sheets in general -- I'd started writing tests for sheets, but we're also looking at a new sheet design (which we're looking at doing in Vueport). No sense doing work we're eventually tossing out :)

src/module/actor/entity.js
```diff
@@ -33,6 +33,15 @@ export class OseActor extends Actor {
     return super.createEmbeddedDocuments(embeddedName, data, context);
   }
 
+  skipRollDialogCheck(options) {
```
wyrmisis, line 36: I think this ought to not be public. We could...
* mark this as private (with a pound sign in front of the name, like we're doing in the data models)
* mark it as static (since it's not using anything from the class)
* move it out of the class and put it in a collection of helper methods.

Thoughts?

src/e2e/actor/character.e2e.test.js
```diff
@@ -5,7 +5,21 @@ export const options = {
   displayName: 'The Character Sheet'
 };
 
-export default ({before, beforeEach, after, describe, it, expect, ...context}) => {
+// TODO: Write Monster roll tests
```
wyrmisis, line 8: This file is meant for tests for the Character actor type. We'll want a separate file for Monsters.

src/module/actor/entity.js
bakbakbakbakbak: Well that is certainly a good point :D

src/module/actor/entity.js
bakbakbakbakbak: I tried marking it as private, but couldn't get it to work (my knowledge is very limited on this), I opted for the last option with my latest commit. (Ps. thanks a lot for your patience!)

src/module/behaviourHelpers.js
```diff
@@ -0,0 +1,10 @@
+export const skipRollDialogCheck = (event) => {
+    const invertedCtrlBehavior = game.settings.get(game.system.id, "invertedCtrlBehavior");
+
+    console.log('Helper triggered.')
```
wyrmisis, line 4: I think we ought to remove this console.log.

src/module/actor/entity.js
wyrmisis: Thank *you* for your patience on all of my comments, hahaha.

src/module/behaviourHelpers.js
bakbakbakbakbak: Haha indeed :)

wyrmisis (approved):
Looks good!

src/lang/de.json
```diff
@@ -181,6 +181,8 @@
   "OSE.Setting.applyDamageOptionHint": "Auf welche(n) Token bei einem Angriffswurf Schaden angewendet werden soll",
   "OSE.Setting.damageSelected": "Ausgewählt",
   "OSE.Setting.damageTarget": "Ziel",
+  "OSE.Setting.InvertedCtrlBehavior": "Roll umkehren Ctrl/Meta-Taste",
+  "OSE.Setting.InvertedCtrlBehaviorHint": "Kehrt das Verhalten beim Halten von Ctrl/Meta beim Klicken auf eine Rolle um.",
```
anthonyronda, line 185: For future reference, please only commit English strings to source. We use https://crowdin.com/project/ose for translating strings

anthonyronda (approved):
LGTM

anthonyronda:
@all-contributors please add @bakbakbakbakbak for code

src/lang/de.json
bakbakbakbakbak: Oooh I had no idea, thanks for letting me know, and: YAAAY :D

Issue #298, opened 2023-01-16
Asking for confirmation when deleting container with items in it


bakbakbakbakbak:
### New Feature or Changed Behaviour

When deleting a container from the ActorSheet, I would like to be asked for confirmation if I really want to do this to prevent mass-spillage of items into my inventory.

### Why?

Deleting a container by mistake containing many items would result in a massive time loss due to clean-up.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #295, opened 2023-01-15
Unable to delete container if populated by item


bakbakbakbakbak:
### What happened?

Steps to reproduce:
1. Create container
2. Create item
3. Drag item into container
4. Try to delete container

### What is the expected behaviour?

When deleting the container, the items should repopulate the correct inventory category.

### Relevant Errors and Warnings

```shell
`Uncaught (in promise) Error: You must provide an _id for every object in the update data Array.
    _preUpdateDocumentArray foundry.js:12648
    _updateEmbeddedDocuments foundry.js:12614
    update commons.js:6770
    updateDocuments commons.js:6128
    updateEmbeddedDocuments commons.js:6321
    _removeItemFromActor actor-sheet.js:115
    activateListeners actor-sheet.js:537
    jQuery 2`
```

### Additional Support Details

Foundry Virtual Tabletop: Version 10, 10.288
Game System: ose, 1.7.3
Active Modules: 0
Performance Mode: 2

OS: Linux x86_64
Client: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0
GPU: Radeon R9 200 Series
Max Texture Size: 8192

No viewed scene

Actors: 1 | Items: 0 | Journal: 0 | Rollable Tables: 0 |
Playlists: 0 | Compendium Packs: 0 | Chat Messages: 2

### Foundry VTT Core Version

V10

### Old-School Essentials Core Version

1.7.3

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #299, opened 2023-01-16
[enhancement] Container delete confirmation

bakbakbakbakbak:
Added a confirmation dialog that proceeds with deleting a container if user selects "Ok".

This only shows when deleting a container that is containing items.

### DIFF
--- a/src/module/actor/actor-sheet.js
+++ b/src/module/actor/actor-sheet.js
@@ -533,8 +533,39 @@ export class OseActorSheet extends ActorSheet {
       item.sheet.render(true);
     });
 
-    html.find(".item-delete").click((event) => {
-      this._removeItemFromActor(event);
+    html.find(".item-delete").click(async (event) => {
+      const item = this._getItemFromActor(event);
+      const itemData = item?.system;
+
+      if (item.type === "container" && itemData.itemIds.length > 0) {
+        // Create dialog
+        let templateData = { types:["ok", "cancel"] },
+          dlg = await renderTemplate(
+            `${OSE.systemPath()}/templates/items/delete-container.html`,
+            templateData
+          );
+  
+        const confirmation = new Dialog({
+          title: game.i18n.localize("OSE.dialog.deleteContainer"),
+          content: dlg,
+          buttons: {
+            ok: {
+              label: game.i18n.localize("OSE.Ok"),
+              icon: '<i class="fas fa-check"></i>',
+              callback: () => {
+                this._removeItemFromActor(event);
+              }
+            },
+            cancel: {
+              icon: '<i class="fas fa-times"></i>',
+              label: game.i18n.localize("OSE.Cancel"),
+            },
+          },
+          default: "cancel"
+        }).render(true);
+      } else {
+        this._removeItemFromActor(event);
+      }
     });
 
     html
--- a/src/module/actor/actor-sheet.js
+++ b/src/module/actor/actor-sheet.js
@@ -535,18 +535,16 @@ export class OseActorSheet extends ActorSheet {
 
     html.find(".item-delete").click((event) => {
       const item = this._getItemFromActor(event);
-      const itemData = item?.system;
 
-      if (item.type === "container" && itemData.itemIds.length > 0) {
-        Dialog.confirm({
-          title: game.i18n.localize("OSE.dialog.deleteContainer"),
-          content: game.i18n.localize("OSE.dialog.confirmDeleteContainer"),
-          yes: () => { this._removeItemFromActor(event); },
-          defaultYes: false
-        })
-      } else {
-        this._removeItemFromActor(event);
-      }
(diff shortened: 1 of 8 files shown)


bakbakbakbakbak:
resolves #298, resolves #295

src/templates/items/delete-container.html
```diff
@@ -0,0 +1,6 @@
+<form class="ose dialog">
+    <div>
+        {{localize 'OSE.dialog.confirmDeleteContainer'}}
+        Are you sure you want to delete the container?
```
wyrmisis, line 4: Should this be localized?

src/module/actor/actor-sheet.js
```diff
@@ -533,8 +533,39 @@ export class OseActorSheet extends ActorSheet {
...
+
+      if (item.type === "container" && itemData.itemIds.length > 0) {
+        // Create dialog
+        let templateData = { types:["ok", "cancel"] },
+          dlg = await renderTemplate(
+            `${OSE.systemPath()}/templates/items/delete-container.html`,
+            templateData
+          );
+  
+        const confirmation = new Dialog({
```
wyrmisis, line 548: You should be able to use `Dialog.confirm` for this: https://foundryvtt.com/api/v10/classes/client.Dialog.html#confirm

wyrmisis (changes requested):
Solid changes, I left some thoughts.

src/templates/items/delete-container.html
bakbakbakbakbak: You're right! I missed to remove my non-localized text!

src/module/actor/actor-sheet.js
bakbakbakbakbak: So much more elegant, thanks, I have now changed to using `Dialog.config` instead.

src/module/actor/actor-sheet.js
```diff
@@ -533,8 +533,20 @@ export class OseActorSheet extends ActorSheet {
       item.sheet.render(true);
     });
 
-    html.find(".item-delete").click((event) => {
-      this._removeItemFromActor(event);
+    html.find(".item-delete").click(async (event) => {
```
wyrmisis, line 536: This arrow function is async, but we're not awaiting anywhere. Should we be awaiting something?

bakbakbakbakbak: No, I missed that when I replaced the custom dialog with the Dialog.confirm - cleaned up in latest commit!

src/module/actor/actor-sheet.js
```diff
@@ -534,7 +534,19 @@ export class OseActorSheet extends ActorSheet {
     });
 
     html.find(".item-delete").click((event) => {
-      this._removeItemFromActor(event);
+      const item = this._getItemFromActor(event);
+      const itemData = item?.system;
+
+      if (item.type === "container" && itemData.itemIds.length > 0) {
```
wyrmisis, line 540: This is more personal preference than anything, but I'm generally a fan of as few branches as possible:

```js
if (item.type !== 'container' || !item?.system?.itemIds?.length)
  return this._removeItemFromActor(event);
  
Dialog.confirm({...});  
```

wyrmisis, line 540: `itemdata = item?.system` is something we'd do if we expect to get here without having a valid item to delete. If so, we might want to add similar checks to the `if` statement here.

wyrmisis (changes requested):
Sorry, just one more change (and a suggestion)

src/module/actor/actor-sheet.js
bakbakbakbakbak: Yes, I realize that I declared `itemData` only as a helper in the if-statement, so it makes no sense to keep given your suggested change above  - fixed in latest commit!

bakbakbakbakbak: I agree - fixed in latest commit!

src/module/actor/actor-sheet.js
```diff
@@ -534,7 +534,17 @@ export class OseActorSheet extends ActorSheet {
     });
 
     html.find(".item-delete").click((event) => {
-      this._removeItemFromActor(event);
+      const item = this._getItemFromActor(event);
+
+      if (item.type !== "container" || !item?.system?.itemIds?.length > 0)
```
wyrmisis, line 539: `item.type` => `item?.type`, since we're checking on the second condition

wyrmisis (changes requested):
Just one last little change!

src/module/actor/actor-sheet.js
bakbakbakbakbak: Indeed, this makes sense!

wyrmisis (approved):
Looks good, thanks for your patience!

anthonyronda:
I was able to test the change in the game system and I'm satisfied it works as intended. Thanks!

PR #301, opened 2023-01-17
feat(i18n): add Catalan and Chinese support to manifest

anthonyronda:


anthonyronda:
We'll also add Swedish

PR #306, opened 2023-01-18
fix: shopping cart now deducts gold

### LINKED ISSUE #282: Shopping Cart Fails to Deduct GP
### What happened?

1. Use OSE CharGen rolling stats and currency
2. Open Cart
3. Move item to inventory
4. Click purchase
5. Coins do not deduct
image

GP should be deducted from the total on the characters sheet.

```shell
No error.
```

Tested on 
M1 Mac Mini 
OS Ventura 13.1
Electron App

### Foundry VTT Core Version

V10

1.6.8

Other

### Code of Conduct

anthonyronda:
- a const wasn't declared properly
- added a missing return when gp item isn't found, otherwise a "not found" console error gets produced
Fixes #282


wyrmisis (approved):
💸

PR #307, opened 2023-01-18
chore: A taste of linting (ose.js, ose.scss.js, == to ===)

wyrmisis:
## Intent

This changeset is meant to show what kinds of changes we can expect to see as we work through linting:
* Changing `==` to `===` and `!=` to `!==`
* Adding some JSDoc frontmatter for file summaries
* Sorting imports
* Changing `let` to `const` (when reasonable)
* Changing from loops to array iteration (where reasonable)
* Avoiding mutation of arguments (where possible)

### DIFF
--- a/src/ose.js
+++ b/src/ose.js
@@ -173,7 +173,7 @@ Hooks.on("renderSidebarTab", async (object, html) => {
 
 Hooks.on("preCreateCombatant", (combat, data, options, id) => {
   let init = game.settings.get(game.system.id, "initiative");
-  if (init == "group") {
+  if (init === "group") {
     OseCombat.addCombatant(combat, data, options, id);
   }
 });
--- a/src/ose.js
+++ b/src/ose.js
@@ -2,28 +2,26 @@
  * @file The entry point for the OSE system
  *       We should handle most of our setup here.
  */
-
-// Import Modules
-import './e2e';
-
 import { OseActorSheetCharacter } from "./module/actor/character-sheet";
-import OseDataModelCharacter from './module/actor/data-model-character';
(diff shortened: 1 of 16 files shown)


src/ose.js
```diff
@@ -1,40 +1,46 @@
...
+import OseDataModelWeapon from './module/item/data-model-weapon';
+import { OseItem } from "./module/item/entity";
+import { OseItemSheet } from "./module/item/item-sheet";
+import * as macros from "./module/macros";
+import * as party from "./module/party";
+import { OsePartySheet } from "./module/party/party-sheet";
+import { preloadHandlebarsTemplates } from "./module/preloadTemplates";
+import * as renderList from "./module/renderList";
+import { registerSettings } from "./module/settings";
+import * as treasure from "./module/treasure";
```
anthonyronda, line 33: imports are alphabetical by filepath? Do we have other options?

Nothing particularly wrong with it from my side, just seems like it's not helpful. Any option for putting all directories first like so?
image

anthonyronda (approved):
I only have the one question, otherwise LGTM

src/ose.js
wyrmisis: I think I got something like this going.

src/ose.js
```diff
@@ -114,27 +114,20 @@ Hooks.once("init", async () => {
...
 Hooks.once("setup", () => {
   // Localize CONFIG objects once up-front
-  [
-    "saves_short",
-    "saves_long",
-    "scores",
-    "armor",
-    "colors",
-    "tags",
-  ].forEach((o) => {
```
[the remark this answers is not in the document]
anthonyronda: (not worth fixing this time)

PR #308, opened 2023-01-18
[fix] Remove options from Monster sheet ability inventory

### LINKED ISSUE #290: Attributes tab not displaying items added by the + button (Monster character sheet)
### What happened?

Adding the "item" from the "+" button on Attributes tab of the monster character sheet sends it to the Inventory tab. 
Added "items" do not display on the Attributes tab under the "+" button.

image
image
image

- clicking the + button on the Attributes tab gives a choice to add abilites or weapons to the abilities and weapons view (adding the item to the monster)
- actual behavior: gives a choice to add any item type, but will only display the added item if it's an ability or weapon

_No response_

_No response_

### Foundry VTT Core Version

V10

1.7.3

_No response_

### Code of Conduct

bakbakbakbakbak:
resolves #290 

Removed the two options according to the discussions in #290 and the second option suggested.


anthonyronda (approved):
LGTM :D

PR #313, opened 2023-01-22
Container streamlining and container bugfixes

### LINKED ISSUE #265: Inventory Drag and Drop Streamline

Inventory itself is a bit cumbersome as you cant drag and drop from items directory into a container it has to go to misc first then dragged into a bag.

There is also not a true way to move things you can only create copies when dragging and dropping from another PC.

### Why?

This just adds time to moving things around.

### Code of Conduct

### LINKED ISSUE #310: Existing Containers prior to 1.7.4 with Contents Will Not Delete
### What happened?

1. Import a Container item from prior to 1.7.4 (github does not support json filetype, but will provide on request.)
2. Add the container to a character sheet.
3. Add items to the container.
4. Attempt to delete - this will fail if you select 'yes'
5. Clear items out of container - this will fail if you select 'yes'

The container should delete.

_No response_

M1 Mac Electron version

### Foundry VTT Core Version

V10

1.7.4

Other

### Code of Conduct

bakbakbakbakbak:
resolves #310 resolves #265

This bug-fix does a few things:
1. It adds CR(UD) tests for containers in `ActorSheets` as discussed in #265 
2. It add code that actually remove items from a containers `itemIds`
4. It reworks  `ActorSheet._removeItemFromActor` to use Item instead of an Event so it can be tested
5. It fixes  `ActorSheet._onDropItem` and makes it correctly drop items eventhough not dropped onto another item
6. It fixes `ActorSheet._onContainerItemAdd` so that it checks if an item exists in actor, otherwise creates it before moving it to a container (fixes #265)


src/module/migration.js
```diff
@@ -0,0 +1,112 @@
...
+    if (actorId && itemIds) {
+        const actor = game.actors.get(actorId);
+        const containerItems = actor.items.reduce((arr, i) => {
+            if ( i.system.containerId === containerId ) 
+                arr.push(i._id);
+
+            return arr;
+        }, []);
+
+        updateData.system = { itemIds: containerItems };
```
wyrmisis, line 108: Is there a way to do this update from the data model?

bakbakbakbakbak: I will investigate!

bakbakbakbakbak: I've looked into the data models but I am not sure how I would approach doing this update from there. Can you be a bit more specific with what you're thinking here?

bakbakbakbakbak: Since I need to access the actors item, I can't use `static migrateData` to perform this update, I did check that at least for the container data model.

anthonyronda:
Regarding whether we'd include this in 1.7.5 or 1.8.0, I think as painful as #310 is, we ought to wait for a minor revision bump for a db migration. I can write mitigation steps for it in release notes

src/module/migration.js
anthonyronda: correct me if I wrong, but I think I saw yesterday there was a path @wyrmisis suggested that @bakbakbakbakbak is still investigating?

bakbakbakbakbak:
> Regarding whether we'd include this in 1.7.5 or 1.8.0, I think as painful as #310 is, we ought to wait for a minor revision bump for a db migration. I can write mitigation steps for it in release notes

I fully agree with this.

src/module/migration.js
bakbakbakbakbak: I looked into it, but was unable to get access to `this.actor` at the `migrateData` step, and to find the `itemIds`, that is unfortunately necessary (as far as I can tell). I checked in the #system-development in the foundry-discord but didn't get any further insight.

Perhaps @wyrmisis has some further insight here?

bakbakbakbakbak: Got it working! The latest changes is migrating containers (checks if the same items that have said containers' `_id` are in the containers `system.itemIds` field, if not it modifies it).

bakbakbakbakbak:
All tests now passes again.

PR #300, opened 2023-01-17
New Crowdin updates

anthonyronda:

PR #319, opened 2023-01-25
Feat/add swedish to manifest add chinese file

anthonyronda:
zh-CN is not auto adding via Crowdin-Github integration in #300 so I have to do it manually

Adds Svenska support in the manifest file (json added in #300)


bakbakbakbakbak (commented):
Can't tell about the Chinese, but the inclusion of Swedish to manifest looks good!

PR #322, opened 2023-01-28
chore: data migration to default img

### LINKED ISSUE #32: Create migration for documents without default icons
I was browsing through my monster list and noticed that most of the ability and weapon icons have been replaced by 'icons/svg/item-bag.svg'. When looking at my actors.db, I can see that these abilities and weapons don't have an 'img' entry. I guess in earlier versions the 'img' entry was only added when the icon had been changed and a default icon was provided for items with empty 'img' entries. Looks like this has been lost along the way.

bakbakbakbakbak:
resolves #32

- Migrate any missing `img` for actors, tokens, and items.

### DIFF
--- a/src/module/item/entity.js
+++ b/src/module/item/entity.js
@@ -24,6 +24,13 @@ export class OseItem extends Item {
     return super.create(data, context);
   }
 
+  static migrateData(source) {
+    if (!source.img && source.type) {
+      source.img = this.defaultIcons(source.type)
+    }
+		return source
+	}
+
   prepareData() {
     super.prepareData();
   }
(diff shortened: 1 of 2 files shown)


src/module/item/entity.js
```diff
@@ -24,6 +24,13 @@ export class OseItem extends Item {
     return super.create(data, context);
   }
 
+  static migrateData(source) {
+    if (!source.img && source.type) {
+      source.img = this.defaultIcons(source.type)
+    }
+		return source
```
bakbakbakbakbak, line 31: Strange format... Looks normal in VSCodium :confused:

bakbakbakbakbak:
Closing this and requests this for 1.8.x after the linting merge

PR #324, opened 2023-01-29
Container streamlining, macro fixes, tests for containers for 1.8.x

### LINKED ISSUE #265: Inventory Drag and Drop Streamline

### LINKED ISSUE #310: Existing Containers prior to 1.7.4 with Contents Will Not Delete

bakbakbakbakbak:
resolves #310 resolves #265

This bug-fix does a few things:
1. It adds CR(UD) tests for containers in `ActorSheets` as discussed in #265 
2. It add code that actually remove items from a containers `itemIds`
4. It reworks  `ActorSheet._removeItemFromActor` to use Item instead of an Event so it can be tested
5. It fixes  `ActorSheet._onDropItem` and makes it correctly drop items eventhough not dropped onto another item
6. It fixes `ActorSheet._onContainerItemAdd` so that it checks if an item exists in actor, otherwise creates it before moving it to a container (fixes #265)
7. Macro fixes cherrypicked from `main`, #317 
8. Linting of the modified files.


bakbakbakbakbak:
Btw we should squash this :sweat_smile:

bakbakbakbakbak:
Closing this to split it into separate PRs for readability

PR #325, opened 2023-01-29
chore: data migration to default img

### LINKED ISSUE #32: Create migration for documents without default icons

bakbakbakbakbak:
Resolves #32 

Adds data migration for actors and items to set default image if missing.

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -8,6 +8,17 @@ export default class OseActor extends Actor {
     this.system.prepareDerivedData?.();
   }
 
+  static migrateData(source) {
+    // Fixing missing img
+    if (!source.img) {
+      source.img = 'icons/svg/mystery-man.svg';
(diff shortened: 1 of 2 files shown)


anthonyronda:
I thought it best to request a review from @wyrmisis if this matches his vision for migrations. It does look clean but once again I'm concerned about the choice of location for the migrations

anthonyronda (approved):
LGTM

Gonna wait for whether @wyrmisis wants to add thoughts before Friday

PR #329, opened 2023-02-01
Chore: Item testing

bakbakbakbakbak:
Adds tests up to item sheet level for items.

I'll see if I can muster the time and energy to refactor this, and make the test more comprehensive for the tags etc. But as far as I can see right now, this is an alright boilerplate if nothing else.

A separate PR with the fixes I've identified due to writing these tests will be tagged to this.


bakbakbakbakbak:
Should I add the tests I've written for the container problems here too perhaps?

PR #334, opened 2023-02-05
chore(test): (almost) all up to sheet level

bakbakbakbakbak:
I know I asked for smaller PRs, but since I got some kind of flow into this the last week, here we go.

I've rewritten most test suites I created. Also added test suites for almost all of the following:
- helpers
- dialog
- item (up to sheet level)
- party (up to sheet level, incl xp)

I have also solved the issues encountered, but will make separate PRs for those. One for helper fixes, and one for the fixes that resolves the issues encountered.

PR #335, opened 2023-02-05
fix: Helper functions discovered faulty during testing

bakbakbakbakbak:
When developing tests in #334, several problems and bugs were discovered, and are fixed in this PR. Will keep in draft until #334 is merged as the fixes are relatively small when that is merged.


bakbakbakbakbak:
This is instead handled separately in #349 #347 & #345.

PR #337, opened 2023-02-05
fix: Always allow attack roll to succeed if not target is selected

### LINKED ISSUE #333: Attack Fails with No Target
### What happened?

1. Create an actor with a weapon
2. Roll attacks with no target.
3. If the attack is 1 or less the attack fails.

The attack roll should always occur regardless of the roll result to hit if there is not a target. The only possible exception would be a 1, which RAW always misses. I propose that this still result in a damage roll with no target, as there is no alternative way for the GM/Players to create a weapon damage roll in chat without a to hit roll preceding it.

```shell
N/A
```

_No response_

### Foundry VTT Core Version

V10

1.7.5

Chrome

### Code of Conduct

bakbakbakbakbak:
Resolves #333 

Built on top of #334 #335 & #336 and will move out of draft once they are merged.


bakbakbakbakbak:
Resolved in #345 instead.

PR #343, opened 2023-02-10
fix: entity-tweaks-tests

bakbakbakbakbak:
Small fix that caused the system not to build on 1.8.x.


anthonyronda (approved):
thanks for the quick fix here

Issue #332, opened 2023-02-05
Way to roll the damage dice manually after see the atack result 


lizieroGUI:
### New Feature or Changed Behaviour

Create a way to roll the damage dice manually after see the atack result.

### Why?

Roll dices is a really fun thing in RPG, i think a lot of people miss this feature.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bakbakbakbakbak:
Is this similar to #333 ? The drafted PR I put up in #337 always rolls the damage die if no target is selected, but doesn't allow it to be rolled individually. If this is not satisfactory, then this needs some more work with the UI etc, which is not my strong suite :)

anthonyronda:
I believe these two issues are connected to the same discussion on Discord. I agree that we could do something like the following:

- a setting to roll separately (this will show as a button in the chat card)
- a separate button for weapon damage rolls only (as mentioned in #333, will need to consult @design)

lizieroGUI:
I think this is perfect, a really good ideia.

Em seg., 6 de fev. de 2023 5:15 PM, Anthony Ronda ***@***.***>
escreveu:

PR #345, opened 2023-02-10
fix(dice): Dice rolling fixes for 1.8.x

### LINKED ISSUE #333: Attack Fails with No Target

### LINKED ISSUE #340: Rolls totalling 1 auto-fail, and rolls totalling 20+ auto-succeed, when it should only be "natural 1s and 20s"
### What happened?

1. Roll a to-hit `1d20` with a natural result of 2 against an AAC of 1
2. Modify the result by -1 (1 vs AAC 1)
3. The roll fails

Also

1. Roll a to-hit `1d20` with a natural result of 19 against an AAC of 21
2. Modify the result by +1 (20 vs AAC 21)
3. The roll succeeds

Attack rolls with a modified result of 1 are allowed to succeed if it hits the target armor class, and rolls with a modified result of 20 are allowed to fail if it misses the target armor class

_No response_

_No response_

### Foundry VTT Core Version

V10

1.7.5

Firefox

### Code of Conduct

### LINKED ISSUE #341: In ascending AC only, natural 1s can succeed and natural 20s can fail
### What happened?

With AAC enabled:

1. Make a weapon with bonus "+20"
2. Roll a natural 1 with no target
3. The attack succeeds

Also
1. Make a weapon with bonus "-20"
2. Roll a natural 20 with no target
3. The attack fails

From RAW: 1s and 20s

"Natural 20 attack rolls always hit. Natural 1 attack rolls always miss."

_No response_

_No response_

### Foundry VTT Core Version

V10

1.7.5

Firefox

### Code of Conduct

bakbakbakbakbak:
resolves #341, resolves #340, resolves #333

- Reworked tests for dice
  - Moved `createMockData` and `createMockAttackData` into tests as they were only used there
  - Updated  `createMockRoll` to properly mock a roll as needed by tests
  - Rewrote the tests for `attackIsSuccess` to better explain functionality
  - Rewrote `digestAttackResult` tests
    - Updated `createMockRoll` calls
    - #340 : Added tests for these cases
    - Reformulated `createMockRoll` calls to calculate values instead of using hard-coded values.
      - Calculations include calculating attack bonus from thac0 score.
  - Rewrote `sendAttackRoll` tests
    - Removed majority of tests that were just checking the chat messages for content
    - Updated the mocking data to more accurately represent actual data from weapon rolling
- Dice helper
  - Fixed `digestResult`
    - Added description
    - Didn't use proper value for `die`
  - Reworked `attackIsSuccess` to make it more understandable. 
    - Added description
    - To use it will take:
      - `roll, target, bonus` for Ascending AC
      - `roll, thac0, ac` for Descending AC
    - Linted the last `if return` statement.
    - #342: Incorporates the changes in said PR.
  - Rewrote `digestAttackResult`
    - Added description
    - #333 : Updated default AC and AAC values to always allow to hit target unless target is selected (except for natural 1s).
    - #341 : Rewrote logic to support Ascending AC & Descending AC utilizing `attackIsSuccess`
    - Properly flags when an Ascending AC roll fails (previously didn't set `result.isFailure`)
  - Added description for `sendAttackRoll`

### DIFF
--- a/src/module/__tests__/helpers-dice.test.ts
+++ b/src/module/__tests__/helpers-dice.test.ts
@@ -430,9 +430,6 @@ export default ({
       rollData.data.roll.dmg = [];
       await OseDice.sendAttackRoll(rollData);
       await waitForInput();
-      await waitForInput();
-      await waitForInput();
-      await waitForInput();
       const notification = ui.notifications?.queue.pop();
       expect(ui.notifications?.queue.length).equal(0);
       expect(notification?.message).equal(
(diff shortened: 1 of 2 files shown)


bakbakbakbakbak:
Marking this as draft until it solves #340 too. Looking into this tomorrow.

src/module/helpers-dice.js
```diff
@@ -82,23 +82,32 @@ const OseDice = {
...
+    // eslint-disable-next-line default-case
     switch (data.roll.type) {
       case "result": {
         if (roll.total === result.target) {
           result.isSuccess = true;
         } else {
           result.isFailure = true;
         }
-
+
```
bakbakbakbakbak, line 110: cleanup

src/module/helpers-dice.js
```diff
@@ -176,44 +202,65 @@ const OseDice = {
     result.target = data.roll.thac0;
     const targetActorData = data.roll.target?.actor?.system || null;
 
-    const targetAc = data.roll.target ? targetActorData.ac.value : 9;
-    const targetAac = data.roll.target ? targetActorData.aac.value : 10;
+    const targetAc = data.roll.target ? targetActorData.ac.value : 20;
+    const targetAac = data.roll.target ? targetActorData.aac.value : -20;
```
anthonyronda, line 206: As I noted in #332 I think this is fine and good for now but doesn't resolve the ultimate UI concern of providing a common sense flow for "no target" combat. We still might as well accept this change as-is because we don't have an enhancement change ready, but I'll be looking forward to designing a new flow in accordance with #332 in the future

src/module/helpers-dice.js
```diff
@@ -176,44 +202,65 @@ const OseDice = {
...
+    } else {
+      result.details = game.i18n.format("OSE.messages.AttackFailure", {
+        bonus: result.target,
+      });
+      result.isFailure = true;
     }
     return result;
   },
 
+  // eslint-disable-next-line jsdoc/require-param
```
anthonyronda, line 245: was this included in an earlier commit and not taken out after including the params in a later commit? Or does this JSDoc comment block somehow not fulfill this rule?

src/module/helpers-dice.js
bakbakbakbakbak: I agree.

src/module/__tests__/helpers-dice.test.ts
```diff
@@ -372,7 +477,6 @@ export default ({
       rollData.data.roll.dmg = [];
       await OseDice.sendAttackRoll(rollData);
       await waitForInput();
-      await waitForInput();
```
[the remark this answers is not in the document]
bakbakbakbakbak: Looking at the test, I'm not sure why I put multiple `waitForInput` here. I removed the multiples and tried running it, and it looks like the tests are just volatile due to timing, which makes sense. This is most likely why I put multiple `waitForInput` before, due to lag.

There might be an idea to rework the tests, to become more deterministic, especially since a lot of them waits for dice rolls, chat card rendering, and sheet rendering.

I'm not really sure what is the appropriate course of action here. I'll spend some time thinking about how to get to the problem.

src/module/helpers-dice.js
bakbakbakbakbak: I couldn't figure it out last time I looked at it, because `param0.speaker` is clearly there. However... There is a missing whitespace between `speaker` and `-`... Fixing that now.

src/module/__tests__/helpers-dice.test.ts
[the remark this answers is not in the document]
anthonyronda: Hmmm, is it ok if I just merge and we'll make a ticket for deterministic waitforinput?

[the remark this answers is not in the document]
bakbakbakbakbak: I would be OK with that, it may be better to have the tests with the potential volatility compared to waiting until this is figured out.

anthonyronda (approved):
LGTM

PR #346, opened 2023-02-11
fix(party): Fix recursive folder adding for 1.8.x

bakbakbakbakbak:
Fixes an issue that was discovered during testing, that if a folder containing a folder that contains an actor. Dragging the head parent folder into the party sheet doesn't add the actor to the party. This should fix all levels of recursion.

### DIFF
--- a/src/module/party/party-sheet.js
+++ b/src/module/party/party-sheet.js
@@ -119,11 +119,11 @@ export default class OsePartySheet extends FormApplication {
 
   _recursiveAddFolder(folder) {
     folder.contents.forEach((actor) => this._addActorToParty(actor));
-    folder.children.forEach((folder) => this._recursiveAddFolder(folder));
+    folder.children.forEach((folder) => this._recursiveAddFolder(folder.folder));
   }
 
   async _onDropFolder(event, data) {
-    if (data.documentName !== "Actor") {
+    if (data.documentName !== "Folder") {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


anthonyronda:
Hmmm, I'm not seeing this getting fixed here. Here's what I'm seeing

behavior in `main` branch with folder > subfolder > actor folder hierarchy:
drag folder: nothing
drag subfolder: actor appears in party overview
drag actor: actor appears in party overview

behavior in #346 
drag folder: nothing
drag subfolder: nothing
drag actor: actor appears in party overview

src/module/party/party-sheet.js
```diff
@@ -119,11 +119,11 @@ export default class OsePartySheet extends FormApplication {
 
   _recursiveAddFolder(folder) {
     folder.contents.forEach((actor) => this._addActorToParty(actor));
-    folder.children.forEach((folder) => this._recursiveAddFolder(folder));
+    folder.children.forEach((folder) => this._recursiveAddFolder(folder.folder));
   }
 
   async _onDropFolder(event, data) {
-    if (data.documentName !== "Actor") {
+    if (data.documentName !== "Folder") {
```
anthonyronda, line 126: I don't think this is correct. The documentName for a folder in the actor subdirectory is always Actor. This line is here to ignore folders from say, the Item subdirectory, or what have you

anthonyronda (changes requested):

src/module/party/party-sheet.js
bakbakbakbakbak: Indeed, and this is what broke the changes :) 

New commit should fix this behaviour!

anthonyronda (approved):
LGTM

PR #347, opened 2023-02-11
fix(macro): Allow macro to be dragged to hotbar for 1.8.x

bakbakbakbakbak:
Macros were not permitted to be dragged to the hotbar as it was restricted to creating item macros. This fixes that behaviour.

Also adds a notification (with localization) that properly warns that the user tries to create a macro from a non-item if that is the case.


anthonyronda (approved):
Beautiful, LGTM

PR #349, opened 2023-02-11
fix(treasure): Make drawTreasure async for 1.8.x

bakbakbakbakbak:
This fixes the deprecation messages for `drawTreasure` using a synchronous call for evaluation of a roll.

### DIFF
--- a/src/module/helpers-treasure.js
+++ b/src/module/helpers-treasure.js
@@ -61,9 +61,9 @@ export const augmentTable = (table, html) => {
  * @param data
  */
 async function drawTreasure(table, data) {
-  const percent = async (chance) => {
+  const percent = (chance) => {
     const roll = new Roll("1d100");
-    await roll.evaluate({ async: true });
+    roll.evaluate({ async: false });
     return roll.total <= chance;
   };
   data.treasure = {};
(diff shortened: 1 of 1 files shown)


anthonyronda:
I'm running into the following unexpected behavior with this change: treasure table selects every row, every draw, as though all rows are 100% chance of being drawn

anthonyronda:
the above behavior is fixed with the following change:

```js
 const percent = (chance) => {
    const roll = new Roll("1d100");
    roll.evaluate({ async: false });
    return roll.total <= chance;
  };
```

basically de-async-ing the roll evaluation. I ran into this as well and it's why I set this particular code aside when doing the V9 deprecation work

HOWEVER! There are no deprecations for synchronous rolls so maybe this is satisfactory?

bakbakbakbakbak:
> the above behavior is fixed with the following change:
> 
> ```js
>  const percent = (chance) => {
>     const roll = new Roll("1d100");
>     roll.evaluate({ async: false });
>     return roll.total <= chance;
>   };
> ```
> 
> basically de-async-ing the roll evaluation. I ran into this as well and it's why I set this particular code aside when doing the V9 deprecation work
> 
> HOWEVER! There are no deprecations for synchronous rolls so maybe this is satisfactory?

Yep! And no deprecation warning in chat :D See new commit!

anthonyronda (approved):
LGTM

PR #351, opened 2023-02-11
fix(dialog): Added class for Entity Tweaks windows for 1.8.x

bakbakbakbakbak:
Putting these two in the same as it is just for convenience for testing purposes.

### DIFF
--- a/src/module/dialog/entity-tweaks.js
+++ b/src/module/dialog/entity-tweaks.js
@@ -7,6 +7,7 @@ export default class OseEntityTweaks extends FormApplication {
   static get defaultOptions() {
     const options = super.defaultOptions;
     options.id = "sheet-tweaks";
+    options.classes = ["sheet-tweaks"];
     options.template = `${OSE.systemPath()}/templates/actors/dialogs/tweaks-dialog.html`;
     options.width = 380;
     return options;


anthonyronda:
Could you describe what for, or provide an example of use, for the ItemTypes? Not opposed just interested. Perhaps the currently commented out Item CRUD tests? Or `sheet-item-tests.ts`?

bakbakbakbakbak:
> Could you describe what for, or provide an example of use, for the ItemTypes? Not opposed just interested. Perhaps the currently commented out Item CRUD tests? Or `sheet-item-tests.ts`?

Mainly because it is being reused in a lot of the tests, and in case we need to add/remove item types later, one location would is simpler to maintain. That was the reasoning of moving this to `testUtils.ts`!

anthonyronda:
I saw it was cherry picked for #356 ? So is this still necessary here?

bakbakbakbakbak:
> I saw it was cherry picked for #356 ? So is this still necessary here?

Indeed, it is not necessary here if/when that gets merged! At the point I submitted this PR I hadn't started #356 yet.

bakbakbakbakbak:
Made this atomic, as part of it is included in #356 , making this just a fix for entity tweaks lacking class like other windows.

anthonyronda (approved):
LGTM

PR #352, opened 2023-02-11
fix(item): Make spendSpell async for 1.8.x

bakbakbakbakbak:
During testing it was found that `spendSpell` needs to be asynchronous as it involves updating the actors, and rendering chat cards.

### DIFF
--- a/src/module/item/entity.js
+++ b/src/module/item/entity.js
@@ -152,9 +152,11 @@ export default class OseItem extends Item {
     });
   }
 
-  spendSpell = async () => {
+  async spendSpell() {
     if (this.type !== "spell")
-      throw new Error("Trying to spend a spell on an item that is not a spell.");
+      throw new Error(
+        "Trying to spend a spell on an item that is not a spell."
+      );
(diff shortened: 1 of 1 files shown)


src/module/item/entity.js
```diff
@@ -152,16 +152,18 @@ export default class OseItem extends Item {
     });
   }
 
-  spendSpell() {
+  spendSpell = async () => {
```
anthonyronda, line 155: why the break with function declaration convention established elsewhere? Did eslint call for it here?

 I thought we were only using `= async ()` for helper function exports not class members, but I could be wrong

anthonyronda (commented):
just the one comment but eager to merge after your response :D

src/module/item/entity.js
bakbakbakbakbak: I think this was done in the heat of the moment while writing all the tests and such! I'll submit a new commit to fix this!

anthonyronda (approved):
LGTM

PR #360, opened 2023-03-03
Symlinker now supports array of dataPaths

### LINKED ISSUE #312: [tools] symlink utility can't handle an array of multiple datapaths in the config file
### What happened?

1. set up a foundryconfig.json file as suggested in contributing.md
2. enter the command npm run link
3. get a console error

symlinks for the ose project get set in multiple "user data" folders

```shell
node:internal/validators:120
    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:371:5)
    at validateString (node:internal/validators:120:11)
    at Object.resolve (node:path:1098:7)
    at Object.<anonymous> (/Users/anthonyronda/red-vtt/ose/tools/symlink.js:25:28)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

_No response_

### Foundry VTT Core Version

None

_No response_

_No response_

### Code of Conduct

bakbakbakbakbak:
resolves #312 

While experimenting with the latest v11 prototype, I figured I'd fix this so I don't have to edit the config more than necessary.

### DIFF
--- a/tools/symlink.js
+++ b/tools/symlink.js
@@ -22,31 +22,41 @@ const foundryConfig = fs.readJSONSync(
   path.resolve(__dirname, "..", "foundryconfig.json")
 );
 
-const linkDirectory = path.resolve(
-  foundryConfig.dataPath,
-  "Data",
-  "systems",
-  foundryConfig.symLinkName
-);
-
 function createSymlink() {
-  console.log(`Linking dist to ${linkDirectory}.`);
-  const systemsDirectory = path.resolve(linkDirectory, "..");
-  if (!fs.pathExistsSync(systemsDirectory)) {
-    throw new Error(
-      `Link directory parent folder "${systemsDirectory}" not found.`
+  foundryConfig.dataPath.forEach((dataPath) => {
+     const linkDirectory = path.resolve(
+      dataPath,
+      "Data",
+      "systems",
+      foundryConfig.symLinkName
     );
-  }
+    
+    console.log(`Linking dist to ${linkDirectory}.`);
+    const systemsDirectory = path.resolve(linkDirectory, "..");
+    if (!fs.pathExistsSync(systemsDirectory)) {
+      throw new Error(
+        `Link directory parent folder "${systemsDirectory}" not found.`
+      );
+    }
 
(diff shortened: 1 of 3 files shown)


anthonyronda:
Appreciate you for this! I started on a solution and forgot all about it because life

amir-arad:
I suggest also fixing the examples in CONTRIBUTING.md (`"symlinkName"` -> `"symLinkName"`)

anthonyronda:
@amir-arad makes a good point, more importantly foundryconfig.json.example needs to be changed too

bakbakbakbakbak:
Alright, updated the CONTRIBUTION.md & also the symlinker.js as it was not having an array of strings as typedef!

tools/symlink.js
```diff
@@ -4,7 +4,7 @@ const yargs = require("yargs");
 const { hideBin } = require("yargs/helpers");
 
 /**
- * @typedef {{dataPath: string; symLinkName: string;}} FoundryConfig
+ * @typedef {{dataPath: string; symLinkName: Array<string>;}} FoundryConfig
```
anthonyronda, line 7: these two are switched

`{{dataPath: Array<string>; symLinkName: string;}} FoundryConfig

anthonyronda (changes requested):
sorry for this! one tiny mistake :)

tools/symlink.js
bakbakbakbakbak: oh, haha just keeping you on the toes ;) but it is fixed in the latest commit!

PR #367, opened 2023-03-17
Fix config examples in CONTRIBUTING.md

amir-arad:
The json configurations in the examples were misleading (both because of #312 and another case sensitivity bug). I changed it to something that works with today's code

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,11 +26,8 @@ Prerequisites: Node.js (v16 is recommended), a familiarity with command Foundry 
 
    ```json
    {
-     "dataPath": [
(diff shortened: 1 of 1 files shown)


bakbakbakbakbak:
I have a PR submitted that fixes the behavior and allowing for linking to several foundry directories. See #360 .

I would advice that over fixing the documentation.

amir-arad:
#360 is the better approach for sure. 

The only benefit in this PR over #360 is that it offers no code changes. If reviewing / testing  is not a problem, I would go with #360.

anthonyronda:
Thanks @amir-arad, with #360 merged I'm going to close this

Issue #59, opened 2022-01-26
EVERYONE: If you've already contributed code or translations, how would you like to be attributed?


anthonyronda:
Anyone whose files, code fragments, docs, or translations has made it to this repo, please give me some information so I can give you proper attribution.

- (required) A name you go by (doesn't have to be your legal name, may be a screen name)
- (optional) An email

If you don't provide an email, I will use your "privacy" email that Github gives all of its users. It looks like `12345+username@noreply.github.com` and it doesn't actually receive mail, it's just an identifier for your code contributions.

From now on I'll be asking every contributor who hasn't already been added for this information

anthonyronda:
I will additionally be using AllContributors, but this is for copyright reasons

engleback:
Just use my name is ok

freohr:
I'm fine with my name and email (both are used in the commit I contributed) in that file

RabidOwlbear:
You can use RabidOwlbear to keep it consistent with my github.

WallaceMcGregor:
I only contributed with Spanish translation and two issues. You can use **WallaceMcGregor** if you wish.

IGALEGOI:
Just use my name is ok
I changed only a very small part, but overall the translation is already good.

anthonyronda:
Everyone above this comment is officially recognized as code authors and translation authors in the latest release 🎉 

If you know someone else who has contributed *before* the switch to github please show them this and ask them to comment as well!

maschill92:
Github information, what's there now is good with me!

bakbakbakbakbak:
My Github username and link to profile is all good for me!

amir-arad:
My Github username and a link to my profile is great

Stew-rt:
Github username and link to profile is all good. Thanks.

anthonyronda:
what would you like your name to be in the credits?
김석환 @eheh031009 on Crowdin for Korean 
little sad kózka @Beteu7 on Crowdin for Polish
Ronner @Ronner231 on Crowdin for Russian

PR #368, opened 2023-03-18
Apply damage to original attack target

### LINKED ISSUE #287: Make damage be dealt to player target token

Save the info of the targeted token when a player attacks, so that when the GM applies damage it deals damage to the token the player originally targeted, removing the need of the GM selecting or targeting the token which damage will be dealt.

### Why?

Not only it removes the responsibility of selecting the right target from the GM, leaving them free to deal with the multitude of things he already needs to, but I've experienced multiple cases where the targeted token didn't update to the gm and they ended up dealing damage to a different token. 

Sometimes it happens by accident, and not very rarely it leads to situations where you dealt say 10 damage the wrong token and now it's dead. But you couldn't remember how many HP left it had before being deducted 10 hit points. Was it 1? Was it 9? Who knows, we don't track negative numbers.

I fully understand that some groups and DMs don't require their players to target before rolling for attacks and prefer to assign damage manually, so this feature would be better done optionally.

### Code of Conduct

amir-arad:
Add the possibility to inflict damage from an attack to the original target of the attack. The original target is determined at the time of the attack, so even if the atytacking player changed target the damage will go to the right actor. 

Protected by `applyDamageOption` (preexisting) feature flag

fixes #287 

image

### DIFF
--- a/src/module/chat.ts
+++ b/src/module/chat.ts
@@ -3,11 +3,11 @@ import { OseActor } from "./actor/entity";
 function canApplyDamage(html: JQuery) {
   if (!html.find('.dice-total').length) return false;
   switch (game.settings.get(game.system.id, "applyDamageOption")) {
-    case "originalTarget":
+    case CONFIG.OSE.apply_damage_options.originalTarget:
       return !!html.find(".chat-target").last().data("id");
-    case "targeted":
+    case CONFIG.OSE.apply_damage_options.targeted:
       return !!game.user?.targets?.size;
-    case "selected":
+    case CONFIG.OSE.apply_damage_options.selected:
       return !!canvas.tokens?.controlled.length;
     default: {
       console.log('unknown setting');
@@ -84,17 +84,17 @@ export const addChatMessageButtons = function (msg: ChatMessage, html: JQuery) {
 function applyChatCardDamage(html: JQuery, multiplier: 1 | -1) {
   const amount = html.find(".dice-total").last().text();
   const dmgTgt = game.settings.get(game.system.id, "applyDamageOption");
-  if (dmgTgt === `originalTarget`) {
+  if (dmgTgt === CONFIG.OSE.apply_damage_options.originalTarget) {
     const victimId = html.find(".chat-target").last().data("id");
     (async () => {
       const actor = ((await fromUuid(victimId || '')) as TokenDocument)?.actor;
       await applyDamageToTarget(actor, amount, multiplier, actor?.name || victimId || 'original target');
     })();
   }
-  if (dmgTgt === `targeted`) {
+  if (dmgTgt === CONFIG.OSE.apply_damage_options.targeted) {
     game.user?.targets.forEach((t) => applyDamageToTarget(t.actor, amount, multiplier, t.name));
   }
-  if (dmgTgt === `selected`) {
+  if (dmgTgt === CONFIG.OSE.apply_damage_options.selected) {
     canvas.tokens?.controlled.forEach((t) => applyDamageToTarget(t.actor, amount, multiplier, t.name));
   }
 }
--- a/src/module/chat.ts
+++ b/src/module/chat.ts
@@ -11,7 +11,10 @@ function canApplyDamage(html: JQuery) {
     case CONFIG.OSE.apply_damage_options.selected:
       return !!canvas.tokens?.controlled.length;
     default: {
-      ui.notifications?.error(`unexpected OSE setting applyDamageOption: ${applyDamageOption}`);
+      ui.notifications?.error(game.i18n.format("OSE.error.unexpectedSettings", {
+        configName: 'applyDamageOption',
+        configValue: applyDamageOption,
+      }));
       return false;
     }
   }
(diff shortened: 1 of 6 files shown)


wyrmisis (changes requested):
I haven't been able to test the changes, but everything here looks reasonable to me. I left some general code quality questions.

src/module/chat.ts
```diff
@@ -1,5 +1,22 @@
 import { OseActor } from "./actor/entity";
 
+function canApplyDamage(html: JQuery) {
+  if (html.find(".dice-total").length)
+    switch (game.settings.get(game.system.id, "applyDamageOption")) {
+      case "originalTarget":
```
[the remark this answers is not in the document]
amir-arad: I can't find a reference for a settings value being derived from `CONFIG.OSE`, a pointer will be welcomed.

IMO The constants here are not magical, they are enum strings. also Typescript matches them against the relevant possible values of `dmgTgt` determined in the settings, where each value has a  defined meaning. changing one of the strings should break TS type check.

src/module/chat.ts
```diff
@@ -62,13 +78,28 @@ export const addChatMessageButtons = function (msg: ChatMessage, html: JQuery) {
...
+ * @param {HTMLElement} html    The chat entry which contains the roll data
  * @param {Number} multiplier   A damage multiplier to apply to the rolled damage.
  * @return {Promise}
  */
-function applyChatCardDamage(roll: JQuery, multiplier: 1 | -1) {
-  const amount = roll.find(".dice-total").last().text();
+function applyChatCardDamage(html: JQuery, multiplier: 1 | -1) {
+  const amount = html.find(".dice-total").last().text();
   const dmgTgt = game.settings.get(game.system.id, "applyDamageOption");
+  if (dmgTgt === `originalTarget`) {
```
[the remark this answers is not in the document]
amir-arad: done

src/module/chat.ts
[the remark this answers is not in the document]
wyrmisis: It's not an exact example, but we're referencing props of `CONFIG.OSE` in the character data model when setting up encumbrance: https://github.com/vttred/ose/blob/main/src/module/actor/data-model-character.js#L21

While Typescript might validate those strings at compile time, I do think that having a single place where these strings are defined, that we then reference when we use them, is a good thing to do. Heck, we can even stick with enums; just export them from somewhere, and refer to them here and in our settings.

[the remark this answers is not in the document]
amir-arad: done

src/module/chat.ts
```diff
@@ -1,5 +1,21 @@
...
+  if (!html.find('.dice-total').length) return false;
+  switch (game.settings.get(game.system.id, "applyDamageOption")) {
+    case CONFIG.OSE.apply_damage_options.originalTarget:
+      return !!html.find(".chat-target").last().data("id");
+    case CONFIG.OSE.apply_damage_options.targeted:
+      return !!game.user?.targets?.size;
+    case CONFIG.OSE.apply_damage_options.selected:
+      return !!canvas.tokens?.controlled.length;
+    default: {
+      console.log('unknown setting');
```
wyrmisis, line 13: I think we should make this more specific, and a UI warning. We *shouldn't* end up here, but if we do, it'd be helpful if a user could tell us how they got here!

src/module/chat.ts
wyrmisis, line 1: @bakbakbakbakbak Just a heads up, I forget if you had tests written for this file or the other stuff in this PR, but this might change things for you.

wyrmisis (commented):
I'm good with the changes here, but I'd appreciate if someone else could have a look!

src/module/chat.ts
bakbakbakbakbak: Thanks for the notice, I'll take a look at this

anthonyronda: probably best to revisit in a later revision of 0.8.x and leave the test as-is until after merging. We're probably expecting a non-zero amount of other surprises after such a large merge

bakbakbakbakbak: Yeah, I am having quite a bit on my plate right now either, and I have noticed that I procrastinate on this check here.

src/templates/chat/roll-attack.html
```diff
@@ -12,8 +12,8 @@ <h2>{{title}}</h2>
             {{/if}}
         </div>
         {{#if result.victim}}
-        <div class="chat-target">
-            vs {{result.victim}}
+        <div class="chat-target" data-id="{{result.victim.document.uuid}}">
+            vs {{result.victim.name}}
```
anthonyronda, line 16: I like this :)

anthonyronda (changes requested):
The PR didn't seem to work as intended (got the new UI error message with each attack attempt) and in debugging I discovered a cause, shown below

src/module/config.ts
```diff
@@ -133,6 +134,11 @@ export const OSE: OseConfig = {
     heavy: "OSE.armor.heavy",
     shield: "OSE.armor.shield",
   },
+  apply_damage_options: {
+    selected : "selected",
+    targeted : "selected",
+    originalTarget : "selected",
+  },
```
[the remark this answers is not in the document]
amir-arad: done

src/module/chat.ts
```diff
@@ -11,7 +11,10 @@ function canApplyDamage(html: JQuery) {
     case CONFIG.OSE.apply_damage_options.selected:
       return !!canvas.tokens?.controlled.length;
     default: {
-      ui.notifications?.error(`unexpected OSE setting applyDamageOption: ${applyDamageOption}`);
+      ui.notifications?.error(game.i18n.format("OSE.error.unexpectedSettings", {
+        configName: 'applyDamageOption',
```
anthonyronda, line 15: Oh nice, good idea :D

anthonyronda (approved):
Confirmed working in Foundry

LGTM

anthonyronda:
Thanks for your contribution @amir-arad 🎉 

@all-contributors please add @amir-arad for code

anthonyronda:
@amir-arad please also take a look at #59 and tell me your preferred name and optional email for entering in the AUTHORS file :)

PR #369, opened 2023-03-20
[Suggestion] a different approach for CONFIG.OSE type definitions

amir-arad:
This is how I prefer to define types where there is a single, constant, value: derive the type from the value. It's all based on the `as const` trick.
this removes duplicated structure, so it's easier to change the code while still allows consumers of the configuration object to be type safe.

This nullifies the `@todo How do I set this type?` comments :)

### DIFF
--- a/src/module/config.ts
+++ b/src/module/config.ts
@@ -14,13 +14,16 @@ export type Color = keyof OseConfig["colors"];
 export type InventoryItemTag = keyof OseConfig["tags"];
 export type EncumbranceOption = keyof OseConfig["encumbranceOptions"];
 
-export const OSE = {
+export const OSE = {  
+  /** Path for system dist */
   systemPath(): string {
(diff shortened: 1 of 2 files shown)


anthonyronda (approved):
LGTM
I see no issues after applying this locally, and it seems to solve the todo's. Nice!

If you enjoy this type of task, the FVTT types team is always looking for more volunteers typing Foundry :D
https://github.com/League-of-Foundry-Developers/foundry-vtt-types

anthonyronda:
@amir-arad I think I did it satisfactorily, but please sanity check my merge commit, where conflicts were merged between apply damage's commit and this one

amir-arad:
> @amir-arad I think I did it satisfactorily, but please sanity check my merge commit, where conflicts were merged between apply damage's branch and this one

LGTM

PR #356, opened 2023-02-17
Chore(test): Actor test suite

bakbakbakbakbak:
This PR introduces tests for Actors up to Sheet level. It is not complete as I've been unable to figure all the tests out, but the test coverage should be quite high with this.

I'm putting this as a draft PR for a few days while I continue linting give a night or two of good rest to simmer.

PR #372, opened 2023-03-28
fix(actor): monster hd rolls

bakbakbakbakbak:
This fixes ~200 tests in #356.


src/module/actor/entity.js
```diff
@@ -282,9 +282,19 @@ export default class OseActor extends Actor {
...
 
     const label = game.i18n.localize(`OSE.roll.hd`);
-    const rollParts = [actorData.hp.hd];
+
+    const rollParts = [];
+
     if (actorType === "character") {
-      rollParts.push(actorData.scores.con.mod * actorData.details.level);
+      const hd = `${actorData.details.level}${actorData.hp.hd.slice(
+        actorData.hp.hd.indexOf("d")
```
anthonyronda, line 290: errors on e.g. `2D6` which is a legal diceterm string (Foundry API lowercases it to `2d6` before rolling)

src/module/actor/entity.js
```diff
@@ -282,9 +282,19 @@ export default class OseActor extends Actor {
...
+      )}`;
+      rollParts.push(hd);
+      rollParts.push(
+        `${actorData.scores.con.mod} * ${actorData.details.level}`
```
anthonyronda, line 294: Please see the discussion here going all the way to the end

https://discord.com/channels/929420928528568390/929511321488203819/1061817746238738482

Particularly here, where the Moldvay Basic wording of the rules is written, and we're given to understand that the OSE rules are meant to emulate the Moldvay Basic rules procedure without repeating them word-for-word

image

There was a later conversation that I can't find, where I *finally* understood the rules

In short: our current implementation is actually incorrect, due to me choosing a different reading of the rules than was apparently intended. We should not be multiplying the con bonus by level

anthonyronda (changes requested):
I am so sorry! all of these tests must have been a big effort to write, but this implementation is based on adapting to our current (wrong) implementation of the game rules. As it is we'll have to completely rethink this. See my explanation below

anthonyronda:
What I think should happen when a character calls rollHitDice based on my understanding of the rules:
```js
if (actorType === "character") {
   rollparts.push(`max(${actorData.hp.hd}+${actorData.scores.con.mod}, 1)`)
....etc
```

anthonyronda:
It's very likely we can port this work over to Character Generator hit dice rolls, given it can generate starting hp for levels >1

anthonyronda:
Don't delete this branch, don't close! I'll make a new enhancement issue as a suggestion for moving this work forward

src/module/actor/entity.js
bakbakbakbakbak: ooh, I'll fix that up

bakbakbakbakbak:
I'll leave the branch as-is, with the capital letter roll fix, and just rebase it to the latest 1.8.x.

PR #377, opened 2023-03-28
fix: removed disabled prop from button tag

### LINKED ISSUE #370: Save button on chat card
### What happened?

Save button in chat card message aren't active. Tried on an existing world with and without module, and on a new test world without anything than the system. No error or warning in the console.
image

The button should be active and roll the save for the selected token.

```shell
Nope, nothing in the console.
```

_No response_

### Foundry VTT Core Version

V10

1.7.6

Chrome, Microsoft Edge

### Code of Conduct

anthonyronda:
Fixes #370 🤷

### DIFF
--- a/src/templates/chat/item-card.html
+++ b/src/templates/chat/item-card.html
@@ -26,7 +26,7 @@
             {{/if}}
 
             {{#if data.save}}
-            <button data-action="save" data-save="{{data.save}}" disabled>
+            <button data-action="save" data-save="{{data.save}}">
                 {{lookup config.saves_long data.save}} - {{localize "OSE.spells.Save"}}
             </button>
             {{/if}}


src/templates/chat/item-card.html
```diff
@@ -26,7 +26,7 @@ <h2>{{item.name}}</h2>
             {{/if}}
 
             {{#if data.save}}
-            <button data-action="save" data-save="{{data.save}}" disabled>
+            <button data-action="save" data-save="{{data.save}}">
```
wyrmisis, line 29: How does this button's event work? Does the selected token save, or the targeted token? Or the token selected/targeted at time of attack?

anthonyronda: Selected token

Issue #366, opened 2023-03-17
Adjust chat card font size for spells,abilities,etc.


krstevenson:
### New Feature or Changed Behaviour

Add a setting to increase the font size of chat card text.  Or apply the Core Setting global font scale to chat cards.

### Why?

Regardless of what Core Setting font size is set to most of the chat card font does not change for anything sent from a character sheet such as spells and ability descriptions.  It is very small and hard to read.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anthonyronda:
I definitely want to leverage the core setting for font size, thanks for pointing that out!

anthonyronda:
@krstevenson please monitor #380 where I'll be sharing screenshots of sheet tweaks relating to this request

there's a few there now

PR #380, opened 2023-03-29
fix: use core font size setting vars

anthonyronda:
Helps address #366 by using the following custom css properties defined in FVTT core. It does not only address chat cards! I'd like to address this as a serious accessibility concern and apply the changes across the system as much as possible.

```css
--font-size-11: 0.6875rem;
--font-size-12: 0.75rem;
--font-size-13: 0.8125rem;
--font-size-14: 0.875rem;
--font-size-16: 1rem;
--font-size-18: 1.125rem;
--font-size-20: 1.25rem;
--font-size-24: 1.5rem;
--font-size-28: 1.75rem;
--font-size-32: 2rem;
--font-size-48: 3rem;
--line-height-12: 0.75rem;
--line-height-16: 1rem;
--line-height-20: 1.25rem;
--line-height-30: 1.875rem;
```

In some cases I was not able to match the existing font sizes and line heights exactly. In some cases I'm concerned that scaling will simply not work. I'm submitting as a draft so we can explore together what additional tweaks can be made

### DIFF
--- a/src/scss/apps.scss
+++ b/src/scss/apps.scss
@@ -38,7 +38,7 @@
     padding: 0;
     .item-entry {
       list-style: none;
-      line-height: 30px;
+      line-height: var(--line-height-30);
       .item-header {
         padding: 0px;
         margin-bottom: 8px;
@@ -53,7 +53,7 @@
           text-indent: 8px;
         }
         .field-short {
-          font-size: 12px;
+          font-size: var(--font-size-12);
           flex-basis: 45px;
           flex-grow: 0;
           text-align: center;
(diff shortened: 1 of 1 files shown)


anthonyronda:
Some example screen captures with the font size setting on a 7 (default is 4)

image
image
image
image
image
image

Some example screen captures with font size on a 10
image
image
image
image

Some of these could certainly be fixed with additional tweaking, but it will be tricky

anthonyronda:
@wyrmisis I think I made a comment about font size in #479 (unless GitHub ate it)

I'd consider all font sizes should use these variables, because a user asked us for accessibility reasons to use Foundry's built in scaling

PR #384, opened 2023-03-31
fix: Clip long item names off with ellipses.

wyrmisis:
## Intent

This changeset clips really long item name text in actor inventories. You can hover over the item's name to see the full name.

image


anthonyronda:
just a sanity check, we already had the hover tooltip?

anthonyronda (approved):
LGTM

PR #386, opened 2023-04-05
fix: isNew() tests now check against saves

anthonyronda:
Previously the tests for isNew() assumed that isNew() was written to check the actor's ability scores, and return true if they were all 0. But isNew() was actually written to check the actor's saves, because this is the best indicator for a "new" Character *and* Monster

```js
  get isNew() {
    return !Object.values(this.saves).reduce(
      (prev, curr) => prev + (parseInt(curr?.value, 10) || 0),
      0
    );
  }
```

I've replaced the tests for Data Model : Monster to check saves, not scores. Data Model : Character now has both, because the saves are autogenerated from the scores, and I think it's useful to check that that's still happening properly.

### DIFF
--- a/src/module/item/entity.js
+++ b/src/module/item/entity.js
@@ -350,6 +350,7 @@ export default class OseItem extends Item {
       actor: this.actor,
       tokenId: token ? `${token.parent.id}.${token.id}` : null,
       item: this._source,
+      itemId: this._source.id,
       data: await this.getChatData(),
       labels: this.labels,
       isHealing: this.isHealing,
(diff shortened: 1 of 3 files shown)


carl-ellis (approved):
This all looks fine, the itemID field seems not entirely related to this issue but still required for compatibility.

src/module/item/entity.js
```diff
@@ -350,6 +350,7 @@ export default class OseItem extends Item {
       actor: this.actor,
       tokenId: token ? `${token.parent.id}.${token.id}` : null,
       item: this._source,
+      itemId: this._source.id,
```
bakbakbakbakbak, line 353: As @carl-ellis mentioned, I don't either get the inclusion of this here. Other than that, the PR looks fine to me!

bakbakbakbakbak (approved):

PR #392, opened 2023-04-09
fix: corrected apply damage settings

anthonyronda:


src/module/config.ts
```diff
@@ -95,9 +95,9 @@ export const OSE = {
     shield: "OSE.armor.shield",
   },
   apply_damage_options: {
-    selected : "selected",
-    targeted : "targeted",
-    originalTarget : "originalTarget",
```
amir-arad, line 100: This will break runtime, as the values in this config are used by `helper-chats.ts` for comparison with the setting(so they need to be equal to the setting value, not the text)

see
https://github.com/vttred/ose/blob/main/src/module/helpers-chat.ts#L16

amir-arad (changes requested):
need to decide if we use the config global objects as enum values, or as settings.

src/module/config.ts
anthonyronda: You're absolutely right, thanks!

PR #393, opened 2023-04-09
fix: apply damage settings

amir-arad:


anthonyronda (approved):
LGTM thanks for catching the problem in my PR and for suggesting a quick fix :)

PR #394, opened 2023-04-09
fix: apply damage to original target was broken

amir-arad:
original target was not passed correctly to chat template


anthonyronda (approved):
LGTM

Also, this appears to fix 4 tests that previously failed in main (although I didn't check which because I tested main first, also I forgot to pull from origin/main so I'm not sure if it's up to date anyway)

PR #396, opened 2023-04-13
fix: containers with phantom items can now delete

anthonyronda:
This is my "elegant" one line fix for #310 

The bug is caused by itemIds that never got removed from container data when the item itself was deleted

There's an argument for creating a migration that removes phantom itemIds from containers rather than using this method. I think we could make an enhancement request for that and remove this?

### DIFF
--- a/src/module/actor/actor-sheet.js
+++ b/src/module/actor/actor-sheet.js
@@ -112,7 +112,9 @@ export default class OseActorSheet extends ActorSheet {
     if (item.type === "container" && item.system.itemIds) {
       const containedItems = item.system.itemIds;
       const updateData = containedItems.reduce((acc, val) => {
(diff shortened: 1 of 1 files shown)


bakbakbakbakbak (commented):
This is actually quite elegant - it makes sense to not push the item back to the actor, if the actor doesn't have the item.

I can't imagine how a container has items that does NOT exist on the actor and it should still be valid to keep the item.

anthonyronda:
I agree.

In case we ever require container validation on any other operations, I suppose the most elegant would be to fix invalid references with a migration step. Not sure which we'd actually need but I can think of a few "nice-to-haves"

bakbakbakbakbak (approved):

bakbakbakbakbak:
I have been thinking and I can't think of any scenarios this wouldn't be appropriate for.

Issue #338, opened 2023-02-06
Individual Initiative/Reroll Each Round is not working.


superpcm:
### What happened?

1. In the OSE Settings, select "Individual initiative" and "Reroll each round".
![image](https://user-images.githubusercontent.com/70666425/217003114-6a215d0b-90aa-43b1-9874-4b2eb6a63374.png)

2. Place two tokens on a map and enter them into combat.
![image](https://user-images.githubusercontent.com/70666425/217003668-f1edb68d-2585-48f7-bbb2-50b74ebaa6da.png)

3. Initial initiative will roll and be presented correctly.
![image](https://user-images.githubusercontent.com/70666425/217083488-7ba50922-80de-4803-895e-902282cd4f46.png)

4. Advance to the next round. Initiative stays the same, no new roll is made for any combatants.
![image](https://user-images.githubusercontent.com/70666425/217083586-ae553284-7035-4749-8005-6724b1b713aa.png)

5. This is the same for any round going forward, initiative does not change.
![image](https://user-images.githubusercontent.com/70666425/217083716-b39d0cfb-9ba4-45b7-bc96-f86632078e97.png)

**Performed on a new world with no modules enabled.**

### What is the expected behaviour?

Every round initiative should be randomized for all combat participants.

### Relevant Errors and Warnings

```shell
None.
```

### Foundry VTT Core Version

V10

### Old-School Essentials Core Version

1.7.5

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anthonyronda:
Fixed in #397

Issue #200, opened 2022-08-05
Don't reroll initiative for round 1 if initiative has already been rolled


anthonyronda:
Note: some groups prefer to roll initiative for their own characters by each clicking the initiative button separately

Expected behavior: 

Either

1. Players and GM may roll PC and NPC initiative separately after starting combat, such as with "Reset each round" selected for rounds subsequent to round 1. 
2. Players and GM may roll PC and NPC initiative separately before starting combat, and upon clicking start combat, the pre-rolled intiative rolls remain for round 1 such as with "keep with each round" selected for rounds subsequent to round 1.

Actual behavior:

Upon clicking Start Combat, all characters are automatically rolled whether there are pre-existing rolls or not.

PR #397, opened 2023-04-14
fix: individual init reroll each round now works

anthonyronda:
Please read #200. 

I unfortunately don't remember how I arrived at the fix in #280 that resulted in bug #338. I think it was quite difficult to find somewhere else to interrupt the unwanted behavior.

The fix here is once again crude, but it's tough because "Start Combat" and "Next Round" are semantically different operations, and we're kind of cornered into using the same initiative function for each.

I think it's worth considering a future refactor such that changes to one combat roll option no longer impact the other two combat round roll options.


wyrmisis (approved):
Looks good to me.

PR #404, opened 2023-04-20
fix: now handling lvl in _createItem correctly

### LINKED ISSUE #401: In Foundry V11, the sheet's create new spell button creates spells without a default level
### What happened?

the Create New Spell button on the spell tab creates spells with level: null 
image

The button on the sheet should instantiate spells with level: 1 by default

_No response_

_No response_

### Foundry VTT Core Version

None

1.8.2

Firefox

### Code of Conduct

anthonyronda:
Fixes #401. 

Also fixes a previously unreported issue (one I noticed), where the Add Spell button on every Level created Level 1 no matter what Level was in that category.

The fix was simple because we are already passing the lvl number with the event dataset. It was just never handled on the actor sheet logic.

### DIFF
--- a/src/module/actor/actor-sheet.js
+++ b/src/module/actor/actor-sheet.js
@@ -402,7 +402,7 @@ export default class OseActorSheet extends ActorSheet {
   _createItem(event) {
     event.preventDefault();
     const header = event.currentTarget;
-    const { treasure, type } = header.dataset;
+    const { treasure, type, lvl } = header.dataset;
     const createItem = (type, name) => ({
       name: name || `New ${type.capitalize()}`,
       type,
@@ -418,6 +418,8 @@ export default class OseActorSheet extends ActorSheet {
     } else {
       const itemData = createItem(type);
       if (treasure) itemData.system = { treasure: true };
+      // when creating a new spell on the character sheet, we need to set the level
+      if (type === "spell") itemData.system = lvl ? { lvl } : {lvl: 1};
       return this.actor.createEmbeddedDocuments("Item", [itemData], {});
     }
   }


src/module/actor/actor-sheet.js
```diff
@@ -418,6 +418,8 @@ export default class OseActorSheet extends ActorSheet {
     } else {
       const itemData = createItem(type);
       if (treasure) itemData.system = { treasure: true };
+      // when creating a new spell on the character sheet, we need to set the level
+      if (type === "spell") itemData.system = lvl ? { lvl } : {lvl: 1};
```
wyrmisis, line 422: I'm surprised this isn't getting some kind of default, weird. I wonder if we could add the default to the data model? At the very least to take the burden of setting a default off of the sheet class.

anthonyronda: Yeah, from 0.3.x to V10 I believe it used the value provided in `template.json` as the default. In V11, it appears to ignore that value. 

Why? At some point I know I heard Atropos say "template.json is going to go away," and I think we're seeing an early effect of that here. I believe now that we're using data models, template.json might as well not exist? That may have been the only relevant bit left

anthonyronda: https://www.youtube.com/watch?v=fYFvDuLvXGk&t=4170s here

anthonyronda: Is there some due diligence I should be doing to push this through? Like looking for more places where defaults are ignored?

anthonyronda: For now let's treat other places where this could happen as a separate ticket. Thanks!

Issue #387, opened 2023-04-05
Linting errors and warnings


anthonyronda:
### What happened?

1. run `npm run build` with or without watch flag
2. Lots of errors and warnings appear
3. The build script still emits a working(?) dist/ folder

### What is the expected behaviour?

Linting errors do not represent build failures but they look like build failures to our new contributors. Even still, we expect for these warnings to be helpful to those contributors so we can encourage them to adhere to a higher code quality standard. Therefore, we should eliminate all linting errors, and most if not all warnings.

### Foundry VTT Core Version

None

### Old-School Essentials Core Version

1.7.7

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anthonyronda:
Grah forgot to link #266 

This comes to us after completing #266

PR #405, opened 2023-04-21
fix: no longer add zeroes to rolls

### LINKED ISSUE #385: Remove +0 from dice roll strings

Whenever the OSE system is putting together an array of dice terms to turn into a dice roll, do not push the dice term to the array if the value is `0`.

The changed behavior will be that the resulting strings will not have `+0` as any dice term

### Why?

It's impacting our ability to test issues with our dice rolls, because "d20 + 0" and "d20" might both be technically correct roll strings, but we can only easily assert 1 string, not every possible correct roll string.

It's also a bit of an eyesore

image

### Code of Conduct

anthonyronda:
Fixes #385.

I'm pretty proud of this one! Rather than place a bandaid, I did some refactoring that I think makes for an overall cleaner rollAttack() function :)

Why did I only minimally touch rollDamage()? I can't figure out where this function is actually used. I think it's used nowhere? It will be useful for enhancements such as #332 so I left it in place for now.

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -1,7 +1,15 @@
-import OseItem from "../item/entity";
-
 import skipRollDialogCheck from "../helpers-behaviour";
 import OseDice from "../helpers-dice";
+import OseItem from "../item/entity";
+
+/**
+ * Used in the rollAttack function to remove zeroes from rollParts arrays
+ *
+ * @param {[]} arr - an array
+ * @returns {[]} - an array
+ */
+const removeFalsyElements = (arr) =>
+  arr.reduce((a, b) => (b ? [...a, b] : a), []);
 
 export default class OseActor extends Actor {
   prepareDerivedData() {
@@ -434,6 +442,20 @@ export default class OseActor extends Actor {
     }
   }
 
+  /**
+   * Constructs the attack roll formula for an attack
+   *
+   * @param {object} attData - the attack data
+   * @param {OseItem} attData.item - the item being used in the attack
+   * @param {OseActor} attData.actor - this actor (at least it should be!)
+   * @param {object} attData.roll - the attack roll data
+   * @param {string} attData.roll.save - the save roll that may be used against the attack
+   * @param {OseActor} attData.roll.target - the target of the attack
+   * @param {object} options - the type of attack and whether to skip the dialog
+   * @param {string} options.type - the type of attack, i.e. melee or missile
+   * @param {boolean} options.skipDialog - whether to skip the dialog
+   * @returns {OseDice.Roll} - the attack roll with completed rollParts and other data
+   */
   rollAttack(attData, options = {}) {
     const data = this.system;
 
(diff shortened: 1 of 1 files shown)


src/module/actor/entity.js
```diff
@@ -437,49 +437,61 @@ export default class OseActor extends Actor {
...
+    if (ascending && data.thac0.bba) {
+      rollParts.push(data.thac0.bba);
     }
+    // for each type of attack, add the Tweaks bonus
+    // and str/dex modifier only if it's non-zero
     if (options.type === "missile") {
       rollParts.push(
-        data.scores.dex.mod.toString(),
-        data.thac0.mod.missile.toString()
+        ...[data.scores.dex.mod, data.thac0.mod.missile].reduce((a, b) => {
```
wyrmisis, line 470: ⛳ 
```js
let attackMods = [];

if (options.type === 'missile')
  attackMods = [data.scores.dex.mod, data.thac0.mod.missile]
else if (options.type === 'melee')
  attackMods = [data.scores.str.mod, data.thac0.mod.melee];
  
rollParts.push(...attackMods.reduce((a, b) => (!b ? a : [...a, b]), []);
```

src/module/actor/entity.js
```diff
@@ -437,49 +437,50 @@ export default class OseActor extends Actor {
...
+      : game.i18n.format("OSE.roll.attacks", {
+          name: this.name,
+        });
+
+    const dmgParts = [
+      // Weapon damage roll value
+      attData.item?.system?.damage ?? "1d6",
+      // Weapon damage bonus
+      attData.item?.system?.bonus,
+    ].reduce((acc, curr) => {
```
wyrmisis, line 453: The reducer here is the same as what's on line 475. We could make that an arrow function constant so we only have to write it once.

anthonyronda:
The arrow function is at the top because that's where ESLint told me to put it
The huge JSDoc is there because I wanted to write a comment for the out-of-context arrow function... then made it a JSDoc... then felt I needed one for the original method I was refactoring...

src/module/actor/entity.js
```diff
@@ -434,52 +442,62 @@ export default class OseActor extends Actor {
...
-      );
-    }
-    if (attData.item && attData.item.system.bonus) {
-      rollParts.push(attData.item.system.bonus);
-    }
-    const thac0 = data.thac0.value;
-    if (options.type === "melee") {
-      dmgParts.push(data.scores.str.mod);
+
+    if (ascending && data.thac0.bba) {
```
wyrmisis, line 480: Since the if/else at line 487 doesn't use curlies, should we remove them here, since we're not doing multiple things in this if block?

anthonyronda: hmmm can I promise to catch it in the #387 PR?

wyrmisis: Hahaha fine, I'll stop nitpicking. :P

wyrmisis (approved):
Clean enough to eat off of. 🚢

src/module/actor/entity.js
anthonyronda: this is not likely the only place where we do this, so I think we'll need to enable an optional linting rule

PR #406, opened 2023-04-21
chore: system.json changes, other superficial changes

anthonyronda:
- I'm now mildly confident that V11 will not blow up in our faces. I'd like to set our maximum compatible version to 11 a bit earlier than last time, because I think that's best for devs and it's going to be fine for users. There will be a warning that they can't migrate their worlds back to V10
- The core library for Simplified Chinese has a different than expected language tag. This corrects the system.json lang key to the correct value
- Package description and sidebar text updated to use some borrowed OSE marketing copy. Trademark has been permitted for use under contract but I'm only just now doing some housekeeping because someone brought it up on Discord recently

### DIFF
--- a/src/templates/chat/license.html
+++ b/src/templates/chat/license.html
@@ -1,23 +1,21 @@
 <div class="ose game-license">
-  <p class="ose game-license">
-    Unofficial Old-School Essentials on Foundry VTT <!-- BEGIN TEXT REQUIRED BY LICENSE -->requires <strong><em>Old-School Essentials</em></strong>.<!-- END TEXT REQUIRED BY LICENSE --> Find the game at <a href="https://necroticgnome.com">Necrotic Gnome's website</a>.
+  <p>
+    Simple, robust rules compatible with decades of B/X and OSR content. <a href="https://necroticgnome.com">See the website for full information about Old-School Essentials.</a>
   </p>
   <p>
-    <!-- BEGIN TEXT REQUIRED BY LICENSE -->
-    This third party product is not affiliated with or approved by Necrotic Gnome.
-
-    Old-School Essentials is a trademark of Necrotic Gnome. The trademark and Old-School Essentials logo are used with permission of Necrotic Gnome, under license. <!-- ADDITIONAL TEXT REQUESTED BY NECROTIC GNOME --> Permission to use Old-School Essentials in the ruleset title granted by Necrotic Gnome. <!-- END ADDITIONAL TEXT -->
-    <!-- END TEXT REQUIRED BY LICENSE -->
+    Old-School Essentials is a trademark of Necrotic Gnome, permitted for use under contract.
   </p>
   <p>
-    Brought to life on Foundry VTT by U~man. Currently maintained by Anthony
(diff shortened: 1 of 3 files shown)


src/templates/chat/license.html
```diff
@@ -1,23 +1,21 @@
 <div class="ose game-license">
-  <p class="ose game-license">
-    Unofficial Old-School Essentials on Foundry VTT <!-- BEGIN TEXT REQUIRED BY LICENSE -->requires <strong><em>Old-School Essentials</em></strong>.<!-- END TEXT REQUIRED BY LICENSE --> Find the game at <a href="https://necroticgnome.com">Necrotic Gnome's website</a>.
+  <p>
+    Simple, robust rules compatible with decades of B/X and OSR content. <a href="https://necroticgnome.com">See the website for full information about Old-School Essentials.</a>
```
wyrmisis, line 3: I have opinions about what text to surround with links, but not enough to block the PR. Thoughts on making just the call to action ("See the website") the link?

src/templates/chat/license.html
```diff
@@ -1,23 +1,21 @@
...
+    <a href="https://github.com/vttred/ose/issues/new">Github issue tracker</a>
+    or by using the Bug Reporter module in Foundry VTT.
   </p>
   <p>
-    <!-- BEGIN TEXT REQUIRED BY LICENSE -->
-    This third party product is not affiliated with or approved by Necrotic Gnome.
-
-    Old-School Essentials is a trademark of Necrotic Gnome. The trademark and Old-School Essentials logo are used with permission of Necrotic Gnome, under license. <!-- ADDITIONAL TEXT REQUESTED BY NECROTIC GNOME --> Permission to use Old-School Essentials in the ruleset title granted by Necrotic Gnome. <!-- END ADDITIONAL TEXT -->
-    <!-- END TEXT REQUIRED BY LICENSE -->
+    Open Source Acknowledgements
```
wyrmisis, line 14: I feel like this should probably be a heading.

wyrmisis (approved):
Looks fine to me, just some subjective stuff.

src/templates/chat/license.html
anthonyronda: I understand that, but I'll have to shrink it with css or go directly to h4/h5/h6

src/templates/chat/license.html
anthonyronda: I lifted that sentence, including the link, exactly from the NG website. But it was neither an expectation from NG nor a goal to be exact, it was just easier to use "vetted" language

In the interest of brevity we can cut it down

anthonyronda:
It occurred to me that Gavin doesn't use the OSR label anywhere (only Basic/Expert), and it was also a personal goal not to use the OSR label anywhere for UFT, so I'd like to retract that bit

anthonyronda:
using h3 because in the settings tab screenshot, General Information and Game Settings are h2's

image

Added padding because we're already inheriting some padding, and I thought this looked more intentional. The top is only weird because of the overflowing span element where the version number is supposed to be

anthonyronda:
Also I want to point out that I don't think UFT is inheriting any of this :)

PR #419, opened 2023-05-02
fix(v11): sidebar license info starts on new line

anthonyronda:
Quick fix in V11testing1's new sidebar layout that keeps the sidebar license info from being in the same row as the system name and version. Going to wait to push until after testing2


anthonyronda:
Same problem is present in V11testing2, so I'm going to go ahead with this fix

PR #444, opened 2023-05-24
fix: move details now save to monsters correctly

anthonyronda:
We had been facing an error where movement details were not getting written to their textbox for a while, and it appeared as though old data may have been wiped. Actually, they were still available in the src of Monster documents, but because their location conflicted with the CharacterMove data model, they were marked as invalid in Foundry and hidden from use.

This changeset moves the movement details storage location to `system.details.movement`, which is more apt for non-gamelogic string data, and creates a data migration where all actors are checked for the old data and moves it to the new location *only* if the new location is empty

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -26,8 +26,10 @@ export default class OseActor extends Actor {
       source.prototypeToken.texture.img = "icons/svg/mystery-man.svg";
     }
     // Fixing missing movement.value by moving it to details.movement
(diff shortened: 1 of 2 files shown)


wyrmisis (approved):
Do we need to update any of our tests? I'm not sure we would since we're doing a migration here.

Otherwise, I'm good.

anthonyronda:
Thanks for the help 💯

Issue #425, opened 2023-05-10
Dragging new Character Sheets does not Link Token to Sheet


seansps:
### What happened?

1. Create a new character sheet
2. Drag the character to the active scene
3. Notice the token is dropped
4. Close the sheet
5. Double click on the token, make a change, close the sheet
6. Open the original sheet, notice that the change was only applied to the token dropped

### What is the expected behaviour?

By default, I'd expect that Character Sheets are always linked back to their Actors. This also makes it so that PCs editing their sheets or setting tokens get updated on the master record.

### Foundry VTT Core Version

V10

### Old-School Essentials Core Version

Latest

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

justinthejaguar:
This seems like a core thing where you are missing a step. Only because you did not mention it, are you linking the data on the actor? 
![Screenshot 2023-05-10 at 8 58 34 AM](https://github.com/vttred/ose/assets/38022481/ab06ef68-e5b8-4488-a161-09ef341ca915)

seansps:
That could be the case, I don't think I ever clicked that.  I am coming from the PF2E system in Foundry, where that is done automatically for users.  So, this is probably more of a suggestion than a bug in this case -- but I think perhaps it should be done automatically for Character Sheets as I never see a case where they'd want it to not behave that way. (Less clicking for setting up new characters too, which have a much shorter life span in OSE!)

Issue #424, opened 2023-05-10
Closing character creation outputs to chat


metodi022:
### What happened?

1. Open the character creation for an actor
2. Close it immediately
3. Observe in the chat there will be a message with empty stats

![image](https://github.com/vttred/ose/assets/10363870/157131d7-99ff-44b8-9cd5-0aab51d82d7c)

### What is the expected behaviour?

If you close the character creation window of an actor, there should be no output in chat.

### Relevant Errors and Warnings

```shell
There are no error messages.
```

### Additional Support Details

Foundry Virtual Tabletop: Version 10, 10.291
Game System: ose, 1.8.4
Active Modules: 16
Performance Mode: 2

OS: Windows NT 10.0; Win64; x64
Client: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/112.0

### Foundry VTT Core Version

V10

### Old-School Essentials Core Version

1.8.4

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Stew-rt:
Individual roll cards (str, int, dex, wis, con, cha and Gold) also render red, "Failure ()" text per roll.

![image](https://github.com/vttred/ose/assets/40280034/559f5f7d-435d-4f32-b695-7769e19a116d)

PR #445, opened 2023-05-27
feat: Issue #425 - Add default "Like Actor Data" for Characters

Stew-rt:
Add default "Link Actor Data" to new Actor's prototypeToken if it's type is character and is not from a Compendium. Also provided two extra options which would be sane but do not cover the Issue, so commented out.

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -94,6 +94,26 @@ export default class OseActor extends Actor {
     return this.system.isNew;
   }
 
+  /**
+   * assign more sane defaults to Actor
+   */
+  async _preCreate(data, options, user) {
+    await super._preCreate(data, options, user);
+    // If this Actor came from a Compendium, do not apply defaults.
+    const sourceId = this.getFlag("core", "sourceId");
+    if (sourceId?.startsWith("Compendium.")) return;
+
+    // Configure prototype token settings
+    const prototypeToken = {};
+    if (this.type === "character")
+      Object.assign(prototypeToken, {
+        // sight: { enabled: true }, // Vision -> Basic Configuration -> Vision Enable
+        actorLink: true, // Identity -> Link Actor Data
+        // disposition: 1, // Identity -> Token Disposition = "Friendly"
+      });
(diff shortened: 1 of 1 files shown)


wyrmisis:
Any chance you could add some testing for this?

Stew-rt:
Sure. Just working on #424 at the mo.. so might be a little bit.

Stew-rt:
I've added tests.. but i'm not sure how this project runs them, sorry if i'm being dumb.

wyrmisis:
Not a problem! The Foundry ecosystem is just weird :)

You'll want to pick up the Quench module. That'll run tests for you.

anthonyronda (changes requested):
Sorry, I ran into some issues with the current changeset

Current code changes the state of a character actor from unlinked to linked when it is copied from a world to a compendium. I would call this unexpected behavior. I believe you could rather check if there *is* a source.id flag before changing the setting, because if there is, it means the user likely wants the token "as-is" without a mutation

There's an additional issue: when duplicating a token, sometimes* the copy is mutated to have Link Actor Token enabled. I believe a token copy should be an exact copy, and that this is also unexpected behavior

* Note: for some odd reason the sourceId is not the token being duplicated. The sourceId is the same as the duplicated token. This means that sometimes it *doesn't* provide what I think is unexpected behavior, if the original token was copied from a compendium. I'll raise this with Foundry, because I think a token copy should have its sourceId flag set to the duplicated token. I don't yet know another way to check if a token is a duplicate

Stew-rt:
Ironically, this implementation is robbed almost verbatim from the official DnD5 system, as my familiarity with the foundry base is limited.

But fair do's, I'll take another look around and try again. 🙂

anthonyronda:
5e? Who plays that? 😆 

Great, thanks, no rush obviously!

anthonyronda:
Noting that the behavior blocking this issue is still awaiting V12

anthonyronda:
Noting that the bugged behavior blocking this issue is still awaiting V12 (it has been pushed several times now)

PR #455, opened 2023-06-04
WIP: feat: Character sheet for v2

wyrmisis:
## Intent

This changeset represents my proposal for a way to build UI elements for the new v2 character sheet.

Of note: it makes use of the [Custom Element](https://developer.mozilla.org/en-US/docs/Web/API/Web_Components/Using_custom_elements) spec and [constructed stylesheets](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet) to encapsulate functionality and styling in components.

Please, please, *please* chime in on this. It's nowhere near done, and that's intentional; I want to get feedback on this approach, because it very much is not something the Foundry dev community uses (outside of stuff like [tagify](https://github.com/yairEO/tagify)).

## What's left?

* Combat tab
  * Displaying AC/AAC per user config
  * Changing fields in the top section of the sheet
    * As Hireling
    * As Advanced Fantasy user
  * Add %bonus to the XP field
  * Add favorite items/abilities functionality
  * All "click to roll" functionality
* All of Abilities tab
* All of Inventory tab
* All of Magic tab
* All of Notes tab


src/templates/actors/character-sheet.html
wyrmisis, line 1: Wait, why the heck is this in here? Hahaha ignore this, I apparently, accidentally, picked up @anthonyronda 's work.

wyrmisis:
Closing in favor of #479

PR #466, opened 2023-07-28
fix: needs to be non-static in order to override

### LINKED ISSUE #460: AC to AAC and vice versa not auto-calculating
On a monster sheet, when switching from AC to AAC or vice versa, the values are not calculated automatically and need to be input separately.

anthonyronda:
Fixes #460 

This function was changed to static as part of a linting pass. We should be checking that we are not changing overrides to static, however, because in this case it caused this `update()` override to never run. The change to static has been reversed. In order to avoid the linter error "do not mutate function properties" we create a clone of the data object property.

This function also returned undefined for all of its conditionals, because (new)`data.system` doesn't exist. The data object's children are flattened into string-keys like this:
image

I switched to this new way of accessing them.

Finally, the linter is complaining about re-using duplicate string literals. This makes no sense in this context, and indeed it should not be triggering because string literals used as object keys are an exception in the SonarSouce/no-duplicate-string rule https://github.com/SonarSource/eslint-plugin-sonarjs/issues/107

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -34,22 +34,24 @@ export default class OseActor extends Actor {
     return source;
   }
 
-  static async update(data, options = {}) {
+  async update(data, options = {}) {
+    const newData = data;
+
     // Compute AAC from AC
     if (data?.ac?.value) {
-      data.aac = { value: 19 - data.ac.value };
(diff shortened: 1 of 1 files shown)


anthonyronda:
The linter is now happy, because the string literals are only used twice each 😆

PR #467, opened 2023-09-13
Encumbrance fix

Typhos:
I noticed that the encumbrance bar seemed to be broken, so I updated the code to return it to its working state. Additionally, I added a quick function to change the color of the bar to indicate how encumbered a character is (green, yellow, orange, red).

### DIFF
--- a/src/module/actor/data-model-character.js
+++ b/src/module/actor/data-model-character.js
@@ -73,6 +73,12 @@ export default class OseDataModelCharacter extends foundry.abstract.DataModel {
       encumbrance: new SchemaField({
         value: new NumberField({ integer: false }),
         max: new NumberField({ integer: false }),
+        pct: new NumberField({ integer: false }),
+        steps: new ObjectField(),
+        atHalfEncumbered: new BooleanField(),
+        atQuarterEncumbered: new BooleanField(),
+        atThreeEighthsEncumbered: new BooleanField(),
+        color: new StringField(),
       }),
       movement: new ObjectField(),
       config: new ObjectField(),
--- a/src/module/actor/data-model-classes/data-model-character-encumbrance.ts
+++ b/src/module/actor/data-model-classes/data-model-character-encumbrance.ts
@@ -96,5 +96,16 @@ export default class OseDataModelCharacterEncumbrance
   get atQuarterEncumbered() {
     return this.value > this.max * (OseDataModelCharacterEncumbrance.encumbranceSteps.quarter / 100) + (this.#delta || 0)
   }
+
+  get color(): string {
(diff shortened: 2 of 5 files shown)


anthonyronda:
thanks, I'll review today :)

Typhos:
I noticed a bug in the encumbrance percentage calculation when I was transferring my fixes manually into my local instance so I could have them for my next session. It looks like the delta value was actually causing the percentages to be incorrect if you use non-standard carrying weights (I have STR modify carrying capacity by 300 per +1). It was working as expected when capacity had a delta of zero.

I removed `this.#delta` from the calculations, which fixed this bug for the bar color feature. I didn't remove `#delta` from the model incase there might be a use for it later in some other feature.

src/module/actor/data-model-character.js
```diff
@@ -73,6 +73,12 @@ export default class OseDataModelCharacter extends foundry.abstract.DataModel {
       encumbrance: new SchemaField({
         value: new NumberField({ integer: false }),
         max: new NumberField({ integer: false }),
+        pct: new NumberField({ integer: false }),
+        steps: new ObjectField(),
+        atHalfEncumbered: new BooleanField(),
+        atQuarterEncumbered: new BooleanField(),
+        atThreeEighthsEncumbered: new BooleanField(),
+        color: new StringField(),
```
wyrmisis, line 81: I'm *pretty sure* the schema is for what's persisted to the DB.  All we persist, as far as I remember, is `max`.

TBH `value` already shouldn't be in there, as it's computed based on the items you're carrying; all we really give the ability to modify is `max`.

src/module/actor/data-model-classes/data-model-character-encumbrance.ts
```diff
@@ -88,13 +89,34 @@ export default class OseDataModelCharacterEncumbrance
...
   get atQuarterEncumbered() {
-    return this.value > this.max * (OseDataModelCharacterEncumbrance.encumbranceSteps.quarter / 100) + (this.#delta || 0)
+    return (
+      this.value >
+      this.max *
+        (OseDataModelCharacterEncumbrance.encumbranceSteps.quarter / 100)
+    );
+  }
+
+  get color(): string {
```
wyrmisis, line 112: Perhaps a spicy take... I think this should be managed on the template, rather than in the encumbrance class. I'm willing to be wrong, though.

wyrmisis (commented):
Some quick thoughts on your changes!

anthonyronda:
We thank you so much for your contribution. It turns out that another contributor was able to come up with a nice solution and answer the review feedback before you did. Thanks a lot though!

PR #496, opened 2023-12-13
Ci/add package release api and build from cache

anthonyronda:
Adds the ability to use this API in CI

https://foundryvtt.com/article/package-release-api/

and re-enables building from cache in CI after this fix

https://github.com/League-of-Foundry-Developers/foundry-vtt-types/pull/2398

(League's Foundry Types has a husky dependency, and husky wants to run scripts upon install–we don't want it to–and it used to error out our build script because it can't find the working directory from its cached directory)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,10 +12,10 @@ jobs:
       - name: Checkout repository to add to workspace
         uses: actions/checkout@v2
 
-      - name: Setup Node.js version 16
+      - name: Setup Node.js version 18
         uses: actions/setup-node@v2
         with:
-          node-version: "16"
+          node-version: "18"
 
       # # Caching to improve build speed
       - name: Enable caching for Node.js modules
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -69,3 +69,23 @@ jobs:
...
+          --data '{
+            "id": "ose",
+            "release": {
+              "version": "${{ github.event.release.tag_name }}",
+              "manifest": "https://github.com/${{github.repository}}/releases/download/${{github.event.release.tag_name}}/system.json",
+              "notes": "https://github.com/https://github.com/${{github.repository}}/releases/tag/${{github.event.release.tag_name}}",
+              "compatibility": {
+                "minimum": "11",
+                "verified": "11",
+                "maximum": "11"
```
wyrmisis, line 88: It'd be cool if these could be dynamically set somehow so we can leave our CI file alone when a new Foundry version comes out.

anthonyronda: You make a good point. This will be fixed before I merge this

anthonyronda:
Using this to preserve environment variables between steps
https://docs.github.com/en/actions/learn-github-actions/variables#about-environment-variables

Issue #382, opened 2023-03-29
New Workflow for "Spend Spell" and "Show Spell"


anthonyronda:
### New Feature or Changed Behaviour

We would like to improve the current interface after a spell or attack is begun. Currently there is a default behavior that we have reason to believe is not optimal for players and GMs to click through

Old Behavior Spend Spell
[screen-recording-2023-03-29-at-111539.webm](https://user-images.githubusercontent.com/1410433/228589865-b032e38f-502e-4ca3-bb66-92d97658c691.webm)

Old Behavior Show Spell
[screen-recording-2023-03-29-at-112150.webm](https://user-images.githubusercontent.com/1410433/228589198-3e75bd36-bc8d-4cd9-bfc8-570e7d461c65.webm)

Suggested New Behavior

**Spend Spell:** Autoroll dice (no button), play Automated Animation, show a single card with dice result and spell description
**Show Spell:** No dice, no animation, show a single card with spell description

We're inviting more suggestions, including mockups

### Why?

This was sparked by the request in #381, which once implemented we noticed shows for both buttons when clicked because both buttons do nearly the same thing. I'd like to take a design-first approach by first asking the community what they think the optimal workflow is before code considerations, because we might not get a well-designed workflow with a code-first, think-later process.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

CLyonsDev:
I think this is a good way to do it. That way players can preview a spell's description without actually casting it, but when they choose to use it, modules like Maestro and Automated Animations can trigger! 👍

PR #497, opened 2024-01-18
Get spell dice rolling

apewall:
This change will make spells roll-able from the character sheet when spells have a roll parameter set.  It also adds the footer and tags to generated cards for both spells and abilities.  Spells will show their descriptions when rolled, but abilities will not.

This brings spells and abilities in line with how weapon attacks rolls function, and should meet the expectations of https://github.com/vttred/ose/issues/382.

<details open>
<summary ="Old spell card">Old spell card</summary>
image
</details>

<details open>
<summary ="New spell card">New spell card</summary>
image
</details>

<details open>
<summary ="Old ability card">Old ability card</summary>
image
</details>

<details open>
<summary ="New ability card">New ability card</summary>
image
</details>


src/templates/chat/roll-result.html
```diff
@@ -8,13 +9,28 @@
...
+        </div>
         <div class="blindable" data-blind="{{data.roll.blindroll}}">
             {{#if result.details}}<div class="chat-details">{{{result.details}}}</div>{{/if}}
             {{#if result.isFailure}}<div class='roll-result roll-fail'><b>{{localize 'OSE.Failure'}}</b> ({{result.target}})
             </div>{{/if}}
             {{#if result.isSuccess}}<div class='roll-result roll-success'><b>{{localize 'OSE.Success'}}</b>
                 ({{result.target}})</div>{{/if}}
             {{#if rollOSE}}<div>{{{rollOSE}}}</div>{{/if}}
+            <div class="card-buttons">
+                {{#if data.save}}
```
wyrmisis, line 23: Thoughts on moving this `#if` up to surround `.card-buttons`?  Unless you expect to see more buttons in this card.

apewall: Since no buttons should exist but saves I think this is reasonable. Resolved in https://github.com/vttred/ose/pull/497/commits/711998f72d185aae1fe4a40074c3efcc2b45f220

src/module/item/entity.js
```diff
@@ -138,32 +138,40 @@ export default class OseItem extends Item {
...
+      description: null,
+      save: itemData.save,
+      properties: this.system.autoTags,
       roll: {
         type,
-        target: data.rollTarget,
-        blindroll: data.blindroll,
+        target: itemData.rollTarget,
+        blindroll: itemData.blindroll,
+        save: itemData.save,
```
Henrik-Bonsmann, line 162: Is there a reason to have both `rollData.save` and `rollData.roll.save`?

src/templates/chat/roll-result.html
```diff
@@ -8,13 +9,28 @@
...
+            {{{data.description}}}
+        </div>
         <div class="blindable" data-blind="{{data.roll.blindroll}}">
             {{#if result.details}}<div class="chat-details">{{{result.details}}}</div>{{/if}}
             {{#if result.isFailure}}<div class='roll-result roll-fail'><b>{{localize 'OSE.Failure'}}</b> ({{result.target}})
             </div>{{/if}}
             {{#if result.isSuccess}}<div class='roll-result roll-success'><b>{{localize 'OSE.Success'}}</b>
                 ({{result.target}})</div>{{/if}}
             {{#if rollOSE}}<div>{{{rollOSE}}}</div>{{/if}}
+            {{#if data.save}}
```
wyrmisis, line 22: Does this come from `rollData.roll.save` or `rollData.save`?

src/module/item/entity.js
apewall: Good catch, roll.save is not necessary and was erroneously included. I've removed it in [a63e2d6](https://github.com/vttred/ose/pull/497/commits/a63e2d681c06c758257605ad2ea2c85b63d26b72)

src/templates/chat/roll-result.html
apewall: This comes from RollData.save, I've removed rollData.roll.save as it is not needed.

wyrmisis (approved):
Looks good to me.

anthonyronda:
Thanks for your review!

@all-contributors add @Henrik-Bonsmann for review

anthonyronda:
@all-contributors add @Henrik-Bonsmann for review

anthonyronda:
@apewall I'm manually testing this PR right now. I'll send the review shortly

apewall:
> @apewall I'm manually testing this PR right now. I'll send the review shortly

Added [b756e01](https://github.com/vttred/ose/pull/497/commits/b756e01d5a0355c515553702c6ca112f42a1c304) to prevent a spell that is not memorized from being cast and display an error to the user.

Henrik-Bonsmann:
> Added [b756e01](https://github.com/vttred/ose/pull/497/commits/b756e01d5a0355c515553702c6ca112f42a1c304) to prevent a spell that is not memorized from being cast and display an error to the user.

While I think that this should be something that's flagged to the user, I don't like the idea of outright preventing the cast. It enforces the use of the automation too much and has the potential to introduce a lot of annoyance for players.

If the preparation was the mistake (instead of the casting), you'd have to open the sheet, navigate to spells, find the correct one, prepare it and then redo (or at least check) all the steps you need to take to cast the spell in the first place.

A warning here is a lot better UX-wise. The best solution for my money would be a "cast anyway" button.

anthonyronda (approved):
This does not fulfill #382 specifically when it comes to Show spell: it still shows an Automated Animations animation for Show spell

I'm happy approving this, as fixing this behavior wasn't specifically what you intended to do, but it doesn't close #382

apewall:
> This does not fulfill #382 specifically when it comes to Show spell: it still shows an Automated Animations animation for Show spell
> 
> I'm happy approving this, as fixing this behavior wasn't specifically what you intended to do, but it doesn't close #382

I only used the discussion in #382 as a reference for what information should  be provided on spell cards when rolled. It's not meant to address the Automation Animation module (Which seems to be in maintenace mode anyway).

anthonyronda:
@all-contributors add apewall for code

Issue #488, opened 2023-11-26
Allow math in SPELL objects


frons79:
### What happened?

![Immagine 2023-11-26 103221](https://github.com/vttred/ose/assets/75011274/49600fe9-e650-4e18-a31b-503cb5f77e01)
I was editing the Hold Monster spell, and I realized it won't allow any math inside **Duration**, **Range** and **Roll** fields.

For instance Hold Monster's duration is _6 + 1/level_ turns and it would be logic to have _9_ turns in the spell card (considering a random level 3 character).

### What is the expected behaviour?

All the math, either with numbers or attributes, possible.

### Foundry VTT Core Version

V11

### Old-School Essentials Core Version

1.9.4

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anthonyronda:
Fixed in #499 to be released tonight

PR #499, opened 2024-01-22
Fix roll formula not being displayed in roll tags

apewall:
This updates the roll tags on abilities and spells so roll formulas can be displayed in the footer cards properly if they include @attributes.

Regarding https://github.com/vttred/ose/issues/488.  Duration and Range could likely be calculated similarly but they would probably need to return totals instead and involve adding additional data for turns/rounds/ft etc.

Before when using a 2nd level character:
image

After:
image


src/module/item/data-model-ability.js
```diff
@@ -21,15 +21,21 @@ export default class OseDataModelAbility extends foundry.abstract.DataModel {
   get #rollTag() {
     if (!this.roll) return null;
 
+    const formulaData = {
```
wyrmisis, line 24: I am irrationally bothered by this code being duplicated. Thoughts on extracting it to a helper method?

I *really* wish extending a base data model class didn't seem to break data models.

apewall: I could consolidate them by adding a method to the dice helper, is this what you are thinking? Would need to import the helper obviously.
```
    return {
      label: `${game.i18n.localize("OSE.items.Roll")} ${OseDice.rollFormula({
        actor: this.parent.actor,
        data: this._source,
      })
        }${rollTarget}`,
```

apewall: I took a stab at adding a new helper in typescript for this in https://github.com/vttred/ose/pull/499/commits/7ecdf48d30fa72471dcb518753ab6e49fa1c276c

src/module/item/data-model-ability.js
```diff
@@ -27,9 +29,11 @@ export default class OseDataModelAbility extends foundry.abstract.DataModel {
         : ` ${CONFIG.OSE.roll_type[this.rollType]}${this.rollTarget}`;
```
wyrmisis, line 29: Guessing we can't move this stuff into the helper, too?

No biggie if not. A little tidier is still tidier.

wyrmisis (approved):
Left a question/request, but I'm good with this if what I'm asking for can't be done.

src/module/item/data-model-ability.js
apewall: I added a new helper named rollTagTarget to generate those as well and removed them from spells since it is not included with spell data anyway.

src/module/item/data-model-ability.js
```diff
@@ -21,15 +23,15 @@ export default class OseDataModelAbility extends foundry.abstract.DataModel {
...
-    const rollTarget =
-      this.rollTarget === undefined
-        ? ""
-        : ` ${CONFIG.OSE.roll_type[this.rollType]}${this.rollTarget}`;
-
     return {
-      label: `${game.i18n.localize("OSE.items.Roll")} ${
-        this.roll
-      }${rollTarget}`,
+      label: `${game.i18n.localize("OSE.items.Roll")} ${OseTags.rollTagFormula({
```
wyrmisis, line 27: Last thing, promise. For legibility, would you mind moving the separate parts of this string (roll label, roll formula, roll target) out into `const`s in this getter?

Then we'd get something like 
```ts
{
  label: `${rollLabel} ${rollFormula}${rollTarget}`
}
```

src/templates/chat/item-card.html
```diff
@@ -25,16 +25,16 @@ <h2>{{item.name}}</h2>
...
             {{#if data.save}}
             <button data-action="save" data-save="{{data.save}}">
                 {{lookup config.saves_long data.save}} - {{localize "OSE.spells.Save"}}
             </button>
             {{/if}}
-
-            {{#if data.roll}}
-            <button data-action="formula">{{ localize "OSE.Roll"}} {{data.roll}} {{#if data.blindroll}}({{localize
-                'OSE.items.BlindRoll'}}){{/if}}</button>
-            {{/if}}
```
anthonyronda, line 37: Same as in #497 (not a problem)

src/module/item/entity.js
```diff
@@ -188,9 +188,8 @@ export default class OseItem extends Item {
 
   _getRollTag(data) {
     if (data.roll) {
-      const roll = `${data.roll}${
-        data.rollTarget ? CONFIG.OSE.roll_type[data.rollType] : ""
-      }${data.rollTarget ? data.rollTarget : ""}`;
```
anthonyronda, line 193: Same as in #497 (not a problem)

src/module/item/entity.js
```diff
@@ -378,6 +377,7 @@ export default class OseItem extends Item {
       hasSave: this.hasSave,
       config: CONFIG.OSE,
     };
+    templateData.rollFormula = new Roll(templateData.data.roll, templateData).formula
```
anthonyronda, line 380: missing semicolon (will merge conflict with #497)

anthonyronda (approved):
LGTM

I noted that there might be a merge conflict while trying to merge after #497, which I can take care of

apewall:
> LGTM
> 
> I noted that there might be a merge conflict while trying to merge after #497, which I can take care of

Sounds good, these ended up related and I forgot to update the formatting on this. :+1:

Issue #498, opened 2024-01-20
Character Inventory not reading external values


Henrik-Bonsmann:
### What happened?

In the character inventory, the "Treasure" value, encumbrance bar and encumbrance break-points are missing. 
Inspecting the element reveals that several handlebars are returning null-values.

In `character-inventory-tab.html`:

```html
    <div class="field-long">
      {{system.treasure}} <i class="fas fa-circle"></i>
    </div>
```
(line 312 f)

```html
<section>
  {{#with @root.system.encumbrance}}
  <div class="encumbrance {{#if encumbered}}encumbered{{/if}}">
    <span class="encumbrance-bar" style="width:{{pct}}%"></span>
    <span class="encumbrance-label">{{value}} / {{max}}</span>
    {{#each steps as |s|}}
    <i class="encumbrance-breakpoint arrow-up" style="left:{{s}}%"></i>
    <i class="encumbrance-breakpoint arrow-down" style="left:{{s}}%"></i>
    {{/each}}
  </div>
  {{/with}}
</section>
```
(line 373 ff)

`{{system.treasure}}`, `{{pct}}` and `{{s}}` return `Null` while  `{{value}}`and `{{max}}` work fine.

### What is the expected behaviour?

Showing the treasure bar with movement break-points and the sum of treasure is vital for keeping bookkeeping low and the game fun and fast paced.

The inventory sheet can be seen working as intended in the guide:
https://ose.vtt.red/images/inventory.png

### Additional Support Details

Foundry: 11.315
OSE: 1.9.4
Client: foundry.exe

### Foundry VTT Core Version

None

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Other

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anthonyronda:
Closed #500

PR #500, opened 2024-02-07
Character sheet inventory fixes

Henrik-Bonsmann:
* fixed encumbrance bar and breakpoints 
* fixed "double weight" issue for treasure weight calculation
* fixed treasure value total in Treasure header
* Unified weight display across Treasure, Misc and Container items to be multiplied with item number

image

### DIFF
--- a/src/module/actor/data-model-character.js
+++ b/src/module/actor/data-model-character.js
@@ -70,10 +70,7 @@ export default class OseDataModelCharacter extends foundry.abstract.DataModel {
       details: new ObjectField(),
       ac: new ObjectField(),
       aac: new ObjectField(),
-      encumbrance: new SchemaField({
-        value: new NumberField({ integer: false }),
-        max: new NumberField({ integer: false }),
-      }),
+      encumbrance: new ObjectField(),
       movement: new ObjectField(),
       config: new ObjectField(),
       initiative: new ObjectField(),
--- a/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
+++ b/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
@@ -19,6 +19,12 @@ export default class OseDataModelCharacterEncumbranceComplete
 
   static templateInventoryRow = "";
 
+  static encumbranceSteps = {
+    quarter: 25,
+    threeEighths: 37.5,
+    half: 50
+  };
+
   /**
    * The machine-readable label for this encumbrance scheme
    */
@@ -49,7 +55,7 @@ export default class OseDataModelCharacterEncumbranceComplete
 
   // eslint-disable-next-line class-methods-use-this
   get steps() {
-    return Object.values(OseDataModelCharacterEncumbrance.encumbranceSteps);
+    return Object.values(OseDataModelCharacterEncumbranceComplete.encumbranceSteps);
   }
 
   get value(): number {
--- a/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
+++ b/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
@@ -49,7 +49,7 @@ export default class OseDataModelCharacterEncumbranceComplete
 
   // eslint-disable-next-line class-methods-use-this
   get steps() {
-    return Object.values(OseDataModelCharacterEncumbranceComplete.encumbranceSteps);
+    return Object.values(OseDataModelCharacterEncumbrance.encumbranceSteps);
   }
 
   get value(): number {
--- a/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
+++ b/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
@@ -19,12 +19,6 @@ export default class OseDataModelCharacterEncumbranceComplete
 
   static templateInventoryRow = "";
 
-  static encumbranceStepsComplete = {
-    quarter: 25,
-    threeEighths: 37.5,
-    half: 50
-  };
-
   /**
    * The machine-readable label for this encumbrance scheme
    */
@@ -55,7 +49,7 @@ export default class OseDataModelCharacterEncumbranceComplete
 
   // eslint-disable-next-line class-methods-use-this
   get steps() {
-    return Object.values(OseDataModelCharacterEncumbranceComplete.encumbranceStepsComplete);
+    return Object.values(OseDataModelCharacterEncumbranceComplete.encumbranceSteps);
   }
 
   get value(): number {
--- a/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
+++ b/src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
@@ -19,7 +19,7 @@ export default class OseDataModelCharacterEncumbranceComplete
 
   static templateInventoryRow = "";
 
-  static encumbranceSteps = {
+  static encumbranceStepsComplete = {
     quarter: 25,
     threeEighths: 37.5,
     half: 50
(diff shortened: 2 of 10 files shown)


Henrik-Bonsmann:
fixes #498

src/templates/actors/partials/character-inventory-tab.html
```diff
@@ -375,7 +375,7 @@ <h4 class="item-name" title="{{item.name}}">{{item.name~}}</h4>
         </div>
         <div class="field-short">
           {{#if (eq @root.config.encumbrance
-          "itembased")}}{{ceil (mult item.system.quantity.value item.system.itemslots)}}{{else}}{{item.system.weight}}{{/if}}{{mult item.system.quantity.value item.system.weight}}
```
wyrmisis, line 378: We're repeating this pattern a few times. I think we should consider doing this on the datamodel.

src/module/actor/data-model-classes/data-model-character-encumbrance-detailed.ts
```diff
@@ -19,6 +19,12 @@ export default class OseDataModelCharacterEncumbranceDetailed
 
   static templateInventoryRow = "";
 
+  static encumbranceSteps = {
```
wyrmisis, line 22: What's your thought process for adding this duplicated static member to the child encumbrance classes?

wyrmisis (changes requested):
Some questions for you.

src/templates/actors/partials/character-inventory-tab.html
Henrik-Bonsmann: I could add new fields for calculated weight, slots and value to the item item.
Though with development on the new character sheet in full swing maybe we can abide a little inelegance on this one for a bit.
Is this expected to be useful in the future? Otherwise this would just add more bloat.

wyrmisis: An `item.system.weightCumulative` getter would be helpful. We'll need to show total weight per Item on whatever our new character sheet is, too!

src/module/actor/data-model-classes/data-model-character-encumbrance-detailed.ts
wyrmisis: > The base class `encumbranceSteps` seems not usable to me, since it now incorporates all 9 possible steps we have over the encumbrance classes (1/4, 3/8, 1/2, 5/8, 3/4, 7/8, 1/3, 5/9, 7/9).

Huh. News to me. I think that came in from @anthonyronda 's item-based encumbrance changes.

> I don't want to change it in case it gets used elsewhere, plus I think it's more elegant to specify in the class itself what the actual steps are.
>
> I assumed I could overwrite the parent member like this and keep it more readable. Would it be better to rename it to `encumbranceStepsDetailed` or is there a way to actually overwrite the field?

Honestly? This feels like a job for JSDoc and comments. Something like:
```js
export default class OseDataModelCharacterEncumbrance
  implements CharacterEncumbrance
{
  ...
  /**
   * Steps at which encumbrance ought to impact Character move speed.
   * 
   * Built-in encumbrance schemes use: 
   * 
   * - `disabled`: None
   * - `basic`: None; this one uses the `significantTreasure` setting.
   * - `detailed`: `quarter`, `threeEighths`, and `half`
   * - `complete`: `quarter`, `threeEighths`, and `half`
   * - `itembased`: 
   *   - Equipped slots use `oneThird`, `fiveNinths`, and `sevenNinths`
   *   - Packed slots use `fiveEighths`, `threeQuarters`, and `sevenEighths`
   */
  static encumbranceSteps = {
    // Detailed/Complete encumbrance steps
    quarter: 25,
    threeEighths: 37.5,
    half: 50,
    // Item-based encumbrance steps
    fiveEighths: 62.5,
    threeQuarters: 75,
    sevenEighths: 87.5,
    oneThird: 33.34,
    fiveNinths: 55.56,
    sevenNinths: 77.78,
  };
  
  ...
}
```

Henrik-Bonsmann: This would definitely clear up the base class a bit, but doesn't really solve the issue of needing only a subset for the derived encumbrances.

Or do I misunderstand something severely?

src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
```diff
@@ -19,6 +19,12 @@ export default class OseDataModelCharacterEncumbranceComplete
 
   static templateInventoryRow = "";
 
+  static encumbranceStepsComplete = {
```
wyrmisis, line 22: I remembered another reason that I'd prefer not to add different `encumbranceSteps` static members to each encumbrance scheme.

Check out [our constants file](https://github.com/vttred/ose/blob/main/src/module/config.ts#L35-L38). We export the selected encumbrance scheme as a constant. I have no clue if someone's built anything that uses this, but we expose it, and it's been exposed for some time.

src/module/actor/data-model-character.js
```diff
@@ -70,10 +70,7 @@ export default class OseDataModelCharacter extends foundry.abstract.DataModel {
       details: new ObjectField(),
       ac: new ObjectField(),
       aac: new ObjectField(),
-      encumbrance: new SchemaField({
-        value: new NumberField({ integer: false }),
-        max: new NumberField({ integer: false }),
-      }),
+      encumbrance: new ObjectField(),
```
wyrmisis, line 73: Would you mind trying something?
```suggestion
      encumbrance: new SchemaField({
        max: new NumberField({ initial: 1600 }),
      }),
```
See if that fixes it, as well. My guess is that we haven't been setting an initial encumbrance max all this time.

If it doesn't work, eh, heck with it, we'll keep it as you have it. If it does, then maybe we add a system setting for default initial value for max carry weight?

src/module/actor/data-model-classes/data-model-character-encumbrance-complete.ts
wyrmisis: The more I think about this whole "look at what item slot encumbrance has done to my sweet boy" thing, the more I wonder if we ought to have a separate move speed class for item-based encumbrance, and select it based on the encumbrance type in use. Item-based encumbrance muddies so much!

Elegant? Maybe not. But it hides our sins, and separates system-specific stuff from what will get ripped out for UFT, once we can (eventually!) get there.

Absolutely not a "this PR" discussion, but something to chew on.

wyrmisis (changes requested):
Sorry to hound you on this one. Let's see if we can't hammer this out together. :)

Maybe there's some stuff we can pull into another PR, if we can isolate the changes that'll fix encumbrance.

src/module/actor/data-model-character.js
wyrmisis: Huh. Fair enough. It was worth a try.

We may want to add a migration to ensure that null/undefined values for `encumbrance.max` are set to `1600` when this goes out. That default gets pulled from `template.json` if we're not specifically setting a schema field up for it.

src/module/item/data-model-item.js
```diff
@@ -19,13 +19,24 @@ export default class OseDataModelItem extends foundry.abstract.DataModel {
...
       quantity: new SchemaField({
-        value: new NumberField({ min: 0, initial: 0 }),
+        value: new NumberField({ min: 0, initial: 1 }),
         max: new NumberField({ min: 0, initial: 0 }),
       }),
       weight: new NumberField({ min: 0, initial: 0 }),
       itemslots: new NumberField({ min: 0, initial: 0 }),
     };
   }
+  get cummulativeWeight(){
```
wyrmisis, line 29: Typo.
```suggestion
  get cumulativeWeight(){
```

wyrmisis (approved):
So clean!

Fix the typo and I'm good with this.

anthonyronda (approved):
LGTM

I'll get started on a PR for initializing the value of encumbrance.max

anthonyronda:
@all-contributors add Henrik-Bonsmann for code

PR #501, opened 2024-02-16
feat: Pug-to-Handlebars Pipeline

wyrmisis:
## Intent

This changeset adds Pug to our workflow.

Pug is a less verbose templating engine that outputs to HTML-like syntax. Thankfully, for our purposes, it'll convert to Handlebars just fine.

### Why the heck?

- HTML is super verbose. It's easy to get lost in a tangle of tags.
- Handlebars templates aren't pre-compiled; they handle the inclusion of partials at runtime. Oftentimes, partials are used for code-sharing and code-splitting. Why foist that cost onto users?


.gitignore
```diff
@@ -3,3 +3,4 @@ foundryconfig.json
 dist
 .DS_Store
 .env
+mprocs.yaml
```
wyrmisis, line 6: I figured, while I'm in here, I might as well ignore the config script for the tool I use to run dev and Foundry from one terminal window.

Henrik-Bonsmann (changes requested):
With a new system in place, I think it might be beneficial to make a note of pug and a link to their documentation somewhere in CONTRIBUTING.md so the on-boarding process for new people is as smooth as possible.

Other than that, I like the look and functionality of pug from what I can see.

wyrmisis:
> With a new system in place, I think it might be beneficial to make a note of pug and a link to their documentation somewhere in CONTRIBUTING.md so the on-boarding process for new people is as smooth as possible.
> 
> Other than that, I like the look and functionality of pug from what I can see.

Absolutely correct. I'm rolling through existing templates to convert them to use Pug right now, but I'll be sure to update our documentation to mention it.

Henrik-Bonsmann (approved):

anthonyronda:
I'll look at this after the bugfix release :)

Issue #503, opened 2024-03-03
THAC0 Providing Incorrect Feedback


justinthejaguar:
### What happened?

Reproduction steps - Keep Descending AC/THAC0 settings
1. Attack without targeting THAC0 19. (no change)
2. Repeat until you get below a ten
3. Receive "Hits AC 9" regardless of missing roll.

### What is the expected behaviour?

This calculation is incorrect. The AC hit for THAC0 should follow the attack matrix in the books/SRD.

### Relevant Errors and Warnings

```shell
https://cdn.discordapp.com/attachments/929772978193956935/1213900434847301692/image.png?ex=65f727e0&is=65e4b2e0&hm=f065fe070e40a9e4b15e26bea8d48ad3d3390632152fde02970d435ce9d9df7f&

A character with THAC0 19 needs a 10 to hit AC 9.
A roll of 5 wouldn't hit anything.
```

### Additional Support Details

Electron - MacOS and Node.js on rpi.

### Foundry VTT Core Version

V11

### Old-School Essentials Core Version

1.10.1

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anthonyronda:
This behavior was purposefully changed in response to #333 and #332, which resulted in the following change

<img width="1169" alt="image" src="https://github.com/vttred/ose/assets/1410433/c20df2f3-ed08-438a-a25c-e52e533de1a5">

compare with @apewall's change

<img width="1165" alt="image" src="https://github.com/vttred/ose/assets/1410433/55bbd0bb-1308-402b-b622-7f44c092880f">

anthonyronda:
Switched to design needed, because clearly there's something wrong with the design of the chat card if people keep changing their minds about whether an attack should hit without a target

PR #504, opened 2024-03-04
Fix: Return proper thac0 attack results when there is no target

apewall:
Resolves https://github.com/vttred/ose/issues/503


wyrmisis (approved):
LGTM, thanks for hopping on this so quickly

anthonyronda:
Please see my comments on #503

src/module/helpers-dice.js
```diff
@@ -208,7 +208,7 @@ const OseDice = {
     result.target = data.roll.thac0;
     const targetActorData = data.roll.target?.actor?.system || null;
 
-    const targetAc = data.roll.target ? targetActorData.ac.value : 20;
+    const targetAc = data.roll.target ? targetActorData.ac.value : 9;
     const targetAac = data.roll.target ? targetActorData.aac.value : -20;
```
anthonyronda, line 212: If you're changing AC, AAC would have to change as well (they always add up to 19)

anthonyronda (changes requested):
As I pointed out in #503, users asked us for attacks without targets to always hit. I think if the expectation is for this to be considered a "bug," we first should address the requests raised in the other issues (#333 and #332) before jumping to a "fix" for this

src/module/helpers-dice.js
apewall: If the idea is to always roll damage when there is no target selected (as opposed to conforming to the THAC0 Table) we should instead just unclamp the THAC0 value.  I think the confusion only comes from the fact it shows as hitting AC 9 regardless of the dice number that was rolled.

https://github.com/vttred/ose/blob/339b156cefd6cc17262a89f95b9d43b2024dd01c/src/module/helpers-dice.js#L235-L241

apewall: https://github.com/vttred/ose/pull/504/commits/f3b7975fde5eb91906879397d0ec9cdceb66016c 
- Unclamped the thac0 result so it will show values outside the original thac0 table.
- Updated success check so if the target is null it will always display attack results if a target is not selected. (Natural 1s still produced failures before.)

https://github.com/vttred/ose/pull/504/commits/f17b7b2afdf7c4923beb856d6720411a05a8a37e
- Changed default target ac and aac values, but these are now irrelevant as they are unused.

PR #517, opened 2024-06-08
v12 compatibility fixes

arcos:
I just started plugging away at this so OSE is usable in v12. I am not sure what the plan is, but hopefully this helps. Fixes dice rolls and makes the character sheets update again.

### DIFF
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -14,8 +14,7 @@ const removeFalsyElements = (arr) =>
 
 export default class OseActor extends Actor {
   prepareDerivedData() {
-    //if (game.version.startsWith("10")) this.system.prepareDerivedData?.();
-    this.system.prepareDerivedData?.();
+    if (game.version.startsWith("10")) this.system.prepareDerivedData?.();
   }
 
   static migrateData(source) {
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -42,7 +42,6 @@ export default class OseActor extends Actor {
       "system.thac0.bba": bbaValue,
       "system.thac0.value": thac0Value,
     } = newData;
-
     // Compute AAC from AC
     if (acValue) {
       newData["system.aac.value"] = 19 - acValue;
@@ -317,8 +316,9 @@ export default class OseActor extends Actor {
 
     const label = game.i18n.localize(`OSE.roll.hd`);
     const rollParts = [actorData.hp.hd];
+    
     if (actorType === "character") {
-      rollParts.push(actorData.scores.con.mod * actorData.details.level);
+      rollParts.push(actorData.scores.con.bonus * actorData.details.level);
     }
 
     const data = {
@@ -550,7 +550,7 @@ export default class OseActor extends Actor {
 
     // Update the Actor
     return this.update({
-      "system.hp.value": Math.clamped(value - amount, 0, max),
+      "system.hp.value": Math.clamp(value - amount, 0, max),
     });
   }
 }
--- a/src/module/actor/entity.js
+++ b/src/module/actor/entity.js
@@ -14,7 +14,8 @@ const removeFalsyElements = (arr) =>
 
 export default class OseActor extends Actor {
   prepareDerivedData() {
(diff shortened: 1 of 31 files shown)


src/module/actor/entity.js
```diff
@@ -14,7 +14,8 @@ const removeFalsyElements = (arr) =>
 
 export default class OseActor extends Actor {
   prepareDerivedData() {
-    if (game.version.startsWith("10")) this.system.prepareDerivedData?.();
+    //if (game.version.startsWith("10")) this.system.prepareDerivedData?.();
+    this.system.prepareDerivedData?.();
```
wyrmisis, line 18: Did this get fixed in v10's patch cycle? v10 doesn't automatically call the data model's `prepareDerivedData`, but v11 (and later, I imagine) does.

src/module/settings.ts
```diff
@@ -67,6 +67,15 @@ const registerSettings = () => {
...
   });
 
+  game.settings.register(game.system.id, "encumbranceItemStrengthMod", {
+    name: game.i18n.localize("OSE.Setting.EncumbranceItemStrengthMod"),
+    hint: game.i18n.localize("OSE.Setting.EncumbranceItemStrengthModHint"),
+    default: false,
+    scope: "world",
+    type: Boolean,
+    config: true,
+  });
```
wyrmisis, line 77: Mind doing this as a separate PR?

src/ose.js
```diff
@@ -32,6 +32,10 @@ import * as treasure from "./module/helpers-treasure";
 
 import "./e2e";
 
+if(typeof Math.clamp !== "function") {
+  Math.clamp = Math.clamped;
+}
```
wyrmisis, line 37: We might as well start a `polyfills.ts` for this kind of stuff.

src/ose.js
```diff
@@ -75,11 +79,11 @@ Hooks.once("init", async () => {
   CONFIG.Actor.documentClass = OseActor;
   CONFIG.Item.documentClass = OseItem;
 
-  CONFIG.Actor.systemDataModels = {
```
wyrmisis, line 78: Is this going to break things for people on v10 or v11? Same with the next set of changes.

src/module/actor/entity.js
arcos: You may know more than me, but under v12 the `prepareDerivedData` method was not being called (the character sheet form didn't update). It appears to work without issue in v11. It's not clear to me what the change is between v11 and v12 though.

src/ose.js
arcos: Reverted, I didn't know that we were keeping v10 support

src/module/actor/entity.js
apewall: It appears that in v12 `prepareDerivedData` is only called automatically if the new TypeDataModel classes are used.
https://github.com/foundryvtt/foundryvtt/issues/10411

Updating the classes to use TypeDataModel should resolve this and make calling prepareDerivedData unnecessary.

https://github.com/vttred/ose/blob/339b156cefd6cc17262a89f95b9d43b2024dd01c/src/ose.js#L78C1-L89C5

arcos: @apewall Thank you! Will this end support for v10 then? Is that ok?

apewall: I believe intention is to only support the latest stable (v12) and maintenance (v11) releases of foundryvtt. 

So it should be okay to drop v10 specific code like this.

system.json
```diff
@@ -5,11 +5,11 @@
...
   "version": "AUTOMATICALLY REPLACED BY GITHUB WORKFLOW ACTION",
   "minimumCoreVersion": "10",
-  "compatibleCoreVersion": "11",
+  "compatibleCoreVersion": "12",
   "compatibility": {
     "minimum": "10",
     "verified": "11",
-    "maximum": "11"
+    "maximum": "12"
   },
```
anthonyronda, line 13: this will probably be changed in a later chore PR

anthonyronda (approved):
LGTM

anthonyronda:
@all-contributors please add arcos for code

PR #523, opened 2024-07-01
Feat: Optionally allow Strength modifier for item-based encumbrance

arcos:
This adds an option to the OSE settings to allow Str mod to apply to carried weight for item-based encumbrance.

image


Rughalt (approved):
LGTM, fixes the encumbrance non-working too. @wyrmisis I think this warrants a look from you too!

src/module/actor/data-model-classes/data-model-character-encumbrance-item-based.ts
```diff
@@ -61,10 +61,16 @@ export default class OseDataModelCharacterEncumbranceItemBased
 
   #packedWeight;
 
+  #weightMod;
+
   // eslint-disable-next-line sonarjs/cognitive-complexity, @typescript-eslint/no-unused-vars
-  constructor(max = 16, items: Item[] = []) {
+  constructor(max = 16, items: Item[] = [], _ = "", strMod = 0) {
```
wyrmisis, line 67: Shouldn't you be able to use `_.scores.str.mod` instead of adding another argument?

wyrmisis (changes requested):

src/module/actor/data-model-character.js
```diff
@@ -25,7 +25,8 @@ export default class OseDataModelCharacter extends foundry.abstract.TypeDataMode
           "significantTreasure"
         ),
         scores: this.scores,
```
wyrmisis, line 27: @arcos This should give you access to the strength mod. The original intent of this object was to give someone who implements their own encumbrance scheme whatever modifiers they might need from the character.

src/module/actor/data-model-classes/data-model-character-encumbrance-item-based.ts
arcos: Thanks!

wyrmisis (approved):
LGTM

Rughalt:
Removed the `.vscode` from `.gitignore` as this is a folder that can and should be shared with project specific settings.

PR #531, opened 2024-07-26
Fix: Item chat cards don't show

### LINKED ISSUE #524: The button to show in the Chat (the eye) does not work
### What happened?

In the Abilities or Spells, we touch the Eye, and nothing happens

The information of the Ability or Spell should be displayed in the chat.

_No response_

_No response_

### Foundry VTT Core Version

V11

1.10.3

_No response_

### Code of Conduct

NekohimeMusou:
Resolves #524.

The method that shows the chat card for an item/ability was referencing `CONST.CHAT_MESSAGE_STYLES` instead of `CONST.CHAT_MESSAGE_TYPES`, which exists. As noted in the issue comments, fixing the reference fixes the behavior. Tested in v11.

### DIFF
--- a/src/module/item/entity.js
+++ b/src/module/item/entity.js
@@ -399,7 +399,7 @@ export default class OseItem extends Item {
     // Basic chat message data
     const chatData = {
       user: game.user.id,
-      type: CONST.CHAT_MESSAGE_STYLES.OTHER,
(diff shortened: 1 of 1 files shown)


Rughalt (approved):
LGTM

PR #533, opened 2024-08-20
Fix: Don't reroll zero init at start of combat

NekohimeMusou:
There's a bit of an edge case in the individual initiative code: if a combatant rolls _exactly_ 0 for their initiative (i.e. due to a penalty) after a combat is created but _before_ combat begins, their initiative will be rerolled at the start of combat because the tracker incorrectly assumes it hasn't been. Checking if the value is nullish, instead of falsy, lets the system distinguish between unrolled init and the number 0.


wyrmisis (approved):
My work on #534 might obviate this PR, but this should still go in so we can cut a final v11 release with fixes for people who need to stay behind.

Issue #535, opened 2024-09-14
Hardcored ose-dev in template path breaks group selector in combat tracker


arcos:
### What happened?

1. Add a bunch of tokens to combat tracker
2. click on the group "Set Combatant Groups" flag icon
3. On an non-dev build you will get an error in the console about not being able to find the template and the group selector modal doesn't open.

### What is the expected behaviour?

The template path is set to the production build path and the group selector opens.

### Relevant Errors and Warnings

```shell
foundry.js:7839 Uncaught (in promise) Error: ENOENT: no such file or directory, open '<path>\systems\ose-dev\dist\templates\apps\combat-set-groups.hbs'
[No packages detected]
    at a.<anonymous> (foundry.js:7839:39)
    at a.value (socket.js:580:13)
    at a.value (socket.js:499:22)
    at U.emit (index.mjs:136:20)
    at manager.js:204:18
(anonymous)	@	foundry.js:7839
value	@	socket.js:580
value	@	socket.js:499
U.emit	@	index.mjs:136
(anonymous)	@	manager.js:204
```

This can be corrected on line 44 of `src\module\combat\combat-set-groups.ts`
```


### Additional Support Details

_No response_

### Foundry VTT Core Version

None

### Old-School Essentials Core Version

_No response_

### Please update your browser regularly. Which (updated) browsers are you seeing the problem on?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Rughalt:
This was fixed by #537

PR #536, opened 2024-09-14
Use correct template path for production build (Issue 535)

arcos:
Fix for https://github.com/vttred/ose/issues/535

### DIFF
--- a/src/module/combat/combat-set-groups.ts
+++ b/src/module/combat/combat-set-groups.ts
@@ -9,7 +9,7 @@ const {
 export default class OSECombatGroupSelector extends HandlebarsApplicationMixin(ApplicationV2) {
   _highlighted;
 
-
+  
   // ===========================================================================
   // APPLICATION SETUP
   // ===========================================================================
@@ -48,7 +48,9 @@ export default class OSECombatGroupSelector extends HandlebarsApplicationMixin(A
 
   // ===========================================================================
   // RENDER SETUP
-@@ -54,48 +51,46 @@ export default class OSECombatGroupSelector extends HandlebarsApplicationMixin(A
+  // ===========================================================================
+
+  async _prepareContext(_options) {
     return {
       groups: colorGroups,
       combatants: game.combat.combatants,
--- a/src/module/combat/combat-set-groups.ts
+++ b/src/module/combat/combat-set-groups.ts
@@ -1,76 +1,79 @@
 import OSE from "../config";
 import { colorGroups } from "./combat-group";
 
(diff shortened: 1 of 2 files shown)


Rughalt (changes requested):
Please remove code reformatting so it's easier to review+

arcos:
> Please remove code reformatting so it's easier to review+

This formatting comes directly from the prettier setup in the codebase. So not sure what I should do here.

arcos:
@Rughalt I reverted the formatting, but we need to have some kind of guideline on how code should be formatted. My IDE picked up the `.prettierrc` from the root of the project and started to auto format (based on my user setting in VS Code), if that's not how it should be done, what is the OSE standard?

wyrmisis:
> @Rughalt I reverted the formatting, but we need to have some kind of guideline on how code should be formatted. My IDE picked up the `.prettierrc` from the root of the project and started to auto format (based on my user setting in VS Code), if that's not how it should be done, what is the OSE standard?

This is generally how it's been done. The only reason it didn't make it in pre-prettified is because my editor didn't already do this (which is a reminder to me to go look at why that is).

src/module/combat/combat-set-groups.ts
```diff
@@ -41,7 +41,7 @@ export default class OSECombatGroupSelector extends HandlebarsApplicationMixin(A
 
   static PARTS = {
     main: {
-      template: `/systems/ose-dev/dist/templates/apps/combat-set-groups.hbs`
+      template: `/systems/${game.system.id}/dist/templates/apps/combat-set-groups.hbs`
```
wyrmisis, line 44: This app finds the template when you do this, right? I think the reason I did this initially was `game.system.id` not being available when this component (but I forgot to swing back around to it and figure something out for it).

wyrmisis (approved):
I'm fine with the changes, provided the combatant group app launches.

src/module/combat/combat-set-groups.ts
arcos: No it doesn't! I thought for sure it would. Ok fixing this

Rughalt:
@wyrmisis @arcos I sat down and thought about prettifying things, and I am all up for it if you do it in two commits in one PR - first, do prettify, and then do the fix. That way we can look at the fix separately from the prettify changes

Rughalt:
I also think I want to fix it the other way; while changing the hardcoded path is good for production it breaks dev-env; what I will do is I will replace all `ose-dev` in the dist files with `ose` during package release procedure

Rughalt:
I added the alternative approach in #537

arcos:
@Rughalt I am going to close this PR, your solution is better.

PR #537, opened 2024-09-15
Add new step in the workflow to replace ose-dev with ose

Rughalt:
This adds new step int he workflow to make sure that all `ose-dev` occurances are replaces `ose` in the dist files


arcos:
Nice, much better solution

arcos (approved):

wyrmisis (approved):
Looks reasonable to me. I can't think of anywhere in the `dist` folder where we'd use `ose-dev` that we genuinely *need* whatever it is to happen for dev-users only.

The only other approach I can think of is to use `.env` or something to set the system ID during the build step, which would be handy for systems that fork off of us. Maybe that's something to look into for the future?

PR #545, opened 2024-10-21
Fix for Treasure Tables not computing successful results.

JustinFreitas:
I noticed that the Treasure Tables were always returning No Results.  I looked into it and apparently the docs are wrong for TextEditor.enrichHTML() because it is returning a promise of a string, not a string.  This just awaits that call, which fixes the Treasure Table rolling.  Yay!

### DIFF
--- a/src/module/helpers-treasure.js
+++ b/src/module/helpers-treasure.js
@@ -123,12 +123,12 @@ export async function rollTreasure(table, options = {}) {
   const html = await renderTemplate(
     `${OSE.systemPath()}/templates/chat/roll-treasure.html`,
     templateData
-    );
+  );
 
   const chatData = {
     content: html,
     // sound: "systems/ose/assets/coins.mp3"
-    };
+  };
 
   const rollMode = game.settings.get("core", "rollMode");
   if (["gmroll", "blindroll"].includes(rollMode))
--- a/src/module/helpers-treasure.js
+++ b/src/module/helpers-treasure.js
@@ -119,25 +119,24 @@ export async function rollTreasure(table, options = {}) {
     });
   }
   
-  setTimeout(async () => {
-    const html = await renderTemplate(
-        `${OSE.systemPath()}/templates/chat/roll-treasure.html`,
-        templateData
-      );
+  await new Promise(resolve => requestAnimationFrame(resolve));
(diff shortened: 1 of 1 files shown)


JustinFreitas:
Huh... only works the first time.  Something is still wrong.  Any ideas?  If not, I can close this down until it works every time.

JustinFreitas:
Wrapping the HTML rendering portion of the method in a setTimeout zero works to correctly populate the chat card every treasure table roll.  I'll see if any of the unit tests were affected.

src/module/helpers-treasure.js
```diff
@@ -118,24 +118,26 @@ export async function rollTreasure(table, options = {}) {
...
-
-  const rollMode = game.settings.get("core", "rollMode");
-  if (["gmroll", "blindroll"].includes(rollMode))
-    chatData.whisper = ChatMessage.getWhisperRecipients("GM");
-  if (rollMode === "selfroll") chatData.whisper = [game.user._id];
-  if (rollMode === "blindroll") chatData.blind = true;
-
-  ChatMessage.create(chatData);
+  
+  setTimeout(async () => {
```
wyrmisis, line 122: Would you mind giving this a try before the `renderTemplate` line? This would replace the `setTimeout`.
```js
await new Promise(resolve => requestAnimationFrame(resolve));
```
The 0ms timeout is probably serving to delay execution of the render and chat card code until the rest of the event loop empties out. Waiting for an animation frame should do similar, but without all the mysticism -- instead, we're just straight up asking the browser to let us know when the UI is ready to receive an update.

wyrmisis (changes requested):
Thanks for having a crack at this -- and sorry it took me a minute to get around to reviewing it.

I have a suggestion I'd appreciate if you could give a go. Let me know if it works. If it doesn't, and the 0ms timeout works, then I'll approve and defer to @Rughalt on getting this merged.

src/module/helpers-treasure.js
JustinFreitas: Ok, it's pushed up.  Thanks again.

wyrmisis (approved):
LGTM

JustinFreitas:
lol

PR #546, opened 2024-11-02
Fix: Slow weapons forgetting their slowness

NekohimeMusou:
Having a Slow weapon equipped currently only slows you down in group initiative if it's the _only_ weapon you have outside of a container. Turns out the isSlow property on the Actor is only true if _all_ of their weapons are both Slow and equipped. Individual initiative likely has the same issue and this change probably fixes that too.


wyrmisis (approved):
Good catch, easy fix. LGTM.

Rughalt (approved):

PR #551, opened 2024-12-19
Party Sheet button fix

JustinFreitas:
Puts the button on the search line before the search icon.  Adjust as necessary.

### DIFF
--- a/src/scss/apps.scss
+++ b/src/scss/apps.scss
@@ -241,6 +241,7 @@
     width: 32px;
     text-align: center;
     line-height: 20px;
+    margin: 4px;
   }
 
   input {
--- a/src/scss/apps.scss
+++ b/src/scss/apps.scss
@@ -239,8 +239,9 @@
 #sidebar #actors .directory-header .header-search {
   .ose-party-sheet {
     width: 32px;
+    height: 25px;
     text-align: center;
-    line-height: 20px;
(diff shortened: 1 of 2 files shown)


wyrmisis:
Mind dropping a screenshot for this change?

JustinFreitas:
> Mind dropping a screenshot for this change?

image

Rughalt (changes requested):
Onbe minor change I would like to request - the button seems to miss left margin and it looks off when taking into account the buttons above it.

JustinFreitas:
> Onbe minor change I would like to request - the button seems to miss left margin and it looks off when taking into account the buttons above it.

Yeah, I can get that figured out.  Was the Party Sheet button on the upper button line before or was it always on the line with the search controls?  Anyway, if you remember, let me know... for now, I'll continue with it on the search line.

Rughalt:
Search line is good, lets just make it nice and tidy CSS wise! Thanks!

src/scss/apps.scss
```diff
@@ -241,6 +241,7 @@
     width: 32px;
     text-align: center;
     line-height: 20px;
+    margin: 4px;
```
JustinFreitas, line 244: Adding in this margin it now looks like this.

image

Rughalt (approved):

PR #556, opened 2025-01-10
Initial setup for devcontainers for OSE project

Rughalt:
This PR adds a simple devcontainer setup that simplifies setting up a development environment.

To use this, you need docker and an editor that supports running devcontainers (or running them locally).


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,24 @@
+{
+	"name": "Node.js & TypeScript",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	"features": {
+		"ghcr.io/devcontainers/features/github-cli:1": {},
+		"ghcr.io/duduribeiro/devcontainer-features/neovim:1": {
+			"version": "nightly"
+		}
```
wyrmisis, line 10: I wonder if there's a way to add LazyVim or Astro to this config?

.devcontainer/setup-foundry.sh
wyrmisis, line 1: We ought to prompt for a license key, too, or have some way to easily pull it from somewhere.

wyrmisis (commented):
A couple questions for a quick review. I'd need to run this before I can approve it. Could you add some notes to the contributing file for how to run it?

.devcontainer/setup-foundry.sh
Rughalt: You can open Foundry then using the provided launch.json (or just by running a command in the devcontainer) and it will ask you for serial number. I don't know how to provide it differently, but I will take a look

